Issue #40, opened 2019-09-11
Choice rule syntax


dhuebner:
- support one of and Choices only in condition blocks.
- condition blocks can have either be a single one of, multiple Choices or multiple Expressions

```
Condition:
    'condition' RosettaNamed? ':' Definable?
        Annotations?
        (constraints += Constraint)+ | (expressions+=RosettaCalcExpression ';')+
;

Constraint:
    Choice | OneOf
;

Choice:
    ('optioinal' | required ?= 'required') 'choice' 'between' attributes += [Attribute | ValidID] ',' attributes += [Attribute | ValidID] (',' attributes += [Attribute | ValidID])*
;

OneOf:
    'one' 'of'
;
```

dhuebner:
Validation:
- [ ] Validate at least two attributes passed to choice

dhuebner:
@jim-h-wang 
I would like to propose `oneOf` or `one-of` instead of `one of` and to drop `between` keyword

dhuebner:
Model Code examples:
```
condition FxSettlementRateSource_choice: required choice settlementRateOption, nonstandardSettlementRate

-//-

condition FxCashSettlement_choice: choice fixing, rateSourceFixing

-//-

condition: oneOf
```

jim-h-wang:
@dhuebner 

> I would like to propose `oneOf` or `one-of` instead of one of and to drop `between` keyword

Let's go with `one-of`, since we have `post-condition` and `assign-output`. Fine to drop the `between`.

> And also drop optional as it's a default

We prefer to keep the explicit qualification, here as this is the preference of our user base.

dhuebner:
@jim-h-wang An other thing where I can't found a case in the existing generator is the handling of multiple constraints `(constraints += Constraint)+`. How should we proceed here?

jim-h-wang:
@dhuebner 

I am thinking of Choices as boolean statements, so we 'and' them together.  For `one-of`, we should add validation to allow only a single use within each `data`.

dhuebner:
@jim-h-wang I wonder how it should work with the current `Validator` API. Especially the `ChoiceRuleFailure` class is restricted to use only one validationMethod type. My understanding is that a validator should be used for only one **named** condition and one set of attributes.

jim-h-wang:
@dhuebner I see your point. We currently don't have a use-case to define 2 Choices within the same Condition block. So for simplicity, let's limit each Condition block to the single Constraint.

[...]

PR #41, opened 2019-09-11
Choice rule migration

dhuebner:
Choice rules like described in #40


com.regnosys.rosetta.lib/src/main/java/model/basictypes.rosetta
```diff
@@ -6,6 +6,7 @@ basicType int
 basicType number
 basicType boolean
 basicType time
+basicType any
```
jim-h-wang: I think we removed the need for the `any` type, by keeping one-of and choice rules as language features. So we should either remove the basic type here or have validation to stop usage outside of the core library.

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -74,7 +74,22 @@ ShortcutDeclaration:
 Condition:
 	'condition' RosettaNamed? ':' Definable?
 		Annotations?
-		(expressions+=RosettaCalcExpression ';')+
+		( constraint = OneOf
```
jim-h-wang: Can we add a validation warning if multiple `one-of`s are used on a `data` object

com.regnosys.rosetta/src/com/regnosys/rosetta/RosettaExtensions.xtend
```diff
@@ -221,4 +230,21 @@ class RosettaExtensions {
 		}
 		return false
 	}
+	
+	def boolean isChoiceRuleCondition(Condition cond) {
+		return cond.constraint !== null
```
jim-h-wang: Something doesn't feel right about this method. We are either testing for existence of Constraints, in which case the name of the function should reflect that; or we are checking for choice rules, in which case the predicate needs to update?

com.regnosys.rosetta.lib/src/main/java/model/basictypes.rosetta
dhuebner: You are right! Just forgot to remove. I will replace it with `empty` later

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
dhuebner: SureI will create a new issue to collect validation rules

com.regnosys.rosetta/src/com/regnosys/rosetta/RosettaExtensions.xtend
dhuebner: It describes the difference between Data rules and Choice rules. That is something you need to consolidate as one validation-concept; over all layers.

jim-h-wang: But `cond.constraint !== null` includes Choice rules as well as OneOfs, so we are treating those 2 things as the same thing in this context. Let's take the naming conversation offline and we can address in subsequence PR.

PR #62, opened 2019-09-19
Drop calculation generation

dhuebner:
We now using only one generator for all `func`s.

Existing ExternalFunction, Function and Calculation Generators are deleted.
Generated classes are changed


com.regnosys.rosetta.tests/src/com/regnosys/rosetta/generator/java/calculation/RosettaCalcExpressionTest.xtend
```diff
@@ -1,6 +1,7 @@
 package com.regnosys.rosetta.generator.java.calculation
 
 import com.google.inject.Inject
+import com.regnosys.rosetta.generator.java.expression.RosettaToJavaExtensions
```
jim-h-wang: unused import?

Issue #70, opened 2019-09-23
Allow only a single expression in 'condition' and remove semi-colon delimiter


jim-h-wang:
All conditions that have multiple Expressions should have expressions `and`ed together.

dhuebner:
@jim-h-wang 
Ready for review #75  and  https://github.com/REGnosys/rosetta-cdm/pull/500

PR #75, opened 2019-09-25
Allow only a single expression in 'condition' and remove ';' #70

dhuebner:


dhuebner:
After that https://github.com/REGnosys/rosetta-cdm/pull/500 should be merged

dhuebner:
merged with #79

PR #87, opened 2019-10-01
Expression generators

dhuebner:
The only change after merging RosettaExpressionJavaGenerator and 
RosettaExpressionJavaGeneratorForFunctions into one ExpressionGenerator is the number literal handling:

`new BigDecimal(0.0) -> BigDecimal.valueOf(0.0)`

Ingest build is passing.


dhuebner:
Mergde

PR #108, opened 2019-11-04
Formatter condition 106

dhuebner:


jim-h-wang:
The `if` statement within the `condition` block should be further indented.

dhuebner:
> The `if` statement within the `condition` block should be further indented.

@jim-h-wang 
We will also need to format the `if` expression and its children, other wise only the `if` part will be indented

PR #163, opened 2020-05-04
namespace hierarchy code-gen

fahd-akbar:


com.regnosys.rosetta/src/com/regnosys/rosetta/generator/java/object/NamespaceHierarchyGenerator.xtend
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Collection
+import java.util.Collections
+import java.util.List
+import java.util.Map
+import java.util.concurrent.atomic.AtomicInteger
+import org.eclipse.xtext.generator.IFileSystemAccess2
+
+class NamespaceHierarchyGenerator {
+
+	val String CDM_NAMESPACE_ROOT = "cdm"
```
minesh-s-patel, line 12: I dont think we can hard code the model name here.

This will mean that all models will have the root of CDM.

com.regnosys.rosetta/src/com/regnosys/rosetta/generator/java/object/NamespaceHierarchyGenerator.xtend
```diff
@@ -0,0 +1,127 @@
...
+	val String ORG_ISDA_CDM_NAMESPACE_ROOT = "org.isda.cdm"
```
minesh-s-patel, line 13: I think this is a special case that we need to cater for that is only a temp.

Maybe deprecate it and add a comment that it should be removed once namespaces are complete?

fahd-akbar: done.

PR #171, opened 2020-05-18
Namespace hierarchy changes v2

fahd-akbar:


com.regnosys.rosetta/src/com/regnosys/rosetta/generator/java/object/NamespaceHierarchyGenerator.xtend
```diff
@@ -98,9 +98,9 @@ class NamespaceHierarchyGenerator {
 			listOfFiles.forEach [ file |
 				var name = switch file {
 					// old file name to remain, but new should change
-					case !file.startsWith("model") && file.endsWith("-enum.rosetta"): "Enum"
-					case !file.startsWith("model") && file.endsWith("-type.rosetta"): "Type"
-					case !file.startsWith("model") && file.endsWith("-func.rosetta"): "Function"
+					case !file.startsWith("model") && file.endsWith("-enum.rosetta"): "enumeration"
```
minesh-s-patel, line 101: Code in DSL should not be used to determine how things are shown on the UI.
The names in the JSON here should be as close to the DSL as possible, and if the UI needs display it differently, then that code should be in the UI.

PR #270, opened 2021-04-21
Fixed issue where the case of an enum dispatch function caused an err…

minesh-s-patel:
…or if it was not uppercase


hugohills-regnosys (approved):

SimonCockx:
@minesh-s-patel This is another really old PR. Is it still of any use, or can it be closed? (just doing a bit of clean-up :))

PR #359, opened 2022-03-16
New dsl list syntax

lolabeis:


documentation/rosetta-modelling-component.md
```diff
@@ -514,53 +555,126 @@ type Confirmation: <"A class to specify a trade confirmation.">
...
+- a *then clause* with the keyword `then` followed by any expression and
+- an optional *else clause* with the keyword `else` followed by any expression
 
-```
-observationEvent -> primitives only-element -> observation
-```
+If the *if clause* evaluates to True, the result of the *then clause* is returned by the conditional expression. If it evaluates to False, the result of the *else clause* is returned if present, else *null* is returned.
 
-This imposes a constraint that the evaluation of the path up to this point returns exactly one value. If it evaluates to [null](#comparison-operators-and-null), an empty list or a list with more than one value then the expression result will be null.
+The type of the expression is the type of the expression contained in the *then clause*. The Rosetta DSL enforces that the type of the *else clause* matches the *then clause*. Multiple *else clauses* can be added by combining `else if` statements ending with a final `else`.
```
hugohills-regnosys, line 585: type and cardinality must match

documentation/rosetta-modelling-component.md
```diff
@@ -514,53 +555,126 @@ type Confirmation: <"A class to specify a trade confirmation.">
...
 
-### Distinct
+### Comparison Operator
 
-The keyword `distinct` can appear after an attribute with multiple cardinality in a Rosetta path. :
+The result type of a comparison operator is always a boolean.
 
+- `=` - Equals. Returns *true* if the left expression is equal to the right expression, otherwise false. Basic types are equal if their values are equal. Two complex rosetta types are equal if all of their attributes are equal, recursing down until all basic typed attributes are compared.
+- `<>` - Does not equal. Returns *false* if the left expression is equal to the right expression, otherwise true.
+- `<`, `<=`, `>=`, `>` - performs mathematical comparisons on the left and right values. Both left and right have to evaluate to numbers or lists of numbers.
```
hugohills-regnosys, line 593: it works with dates too

hugohills-regnosys: actually any comparable type: string, int, number, date

documentation/rosetta-modelling-component.md
```diff
@@ -643,81 +757,96 @@ func GetDrivingLicenceNames: <"Get driver's names from given list of licences.">
...
 
-An expression can refer to a Rosetta Enumeration value using the name of the enumeration type followed by `->` and the name of the enumeration value. E.g. :
+Reduction consists of a set of operations that returns a single value based on elements of a list.
 
-```
-DayOfWeekEnum -> SAT
-```
+- `sum` - returns the sum of the elements of a list of `int` or `number`
+- `min`, `max` - returns the minimum or maximum of a list of comparable elements (`int`, `number`, `date` and `string`)
+- `join` - returns the concatenated values of a list of `string`
```
hugohills-regnosys, line 770: reduce keyword is missing from this list (but is covered below)

hugohills-regnosys (approved):

PR #385, opened 2022-07-27
Refactor pojo builder pattern

hugohills-regnosys:


SimonCockx:
@hugohills-regnosys Is this PR still of any use, or can it be closed? (just doing a bit of clean-up :))

PR #399, opened 2022-09-09
Unknown OSGi execution environment: 'JavaSE-18'

maoo:
Tried to `mvn install` from a fresh project checkout (using both Java 14 and Java 18) and got the following error

```
[ERROR] Internal error: org.eclipse.tycho.core.ee.UnknownEnvironmentException: Unknown OSGi execution environment: 'JavaSE-18'
```

The issue was solved by updating the Tycho plugins to its latest version, `2.7.5`.

Hope this helps!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@ See the Apache License Version 2.0 for the specific language governing permissio
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<maven.compiler.release>8</maven.compiler.release>
 		<!-- Tycho settings -->
-		<tycho-version>2.5.0</tycho-version>
+		<tycho-version>2.7.5</tycho-version>
 		<!-- Define overridable properties for tycho-surefire-plugin IMPORTANT THESE PROPERTIES ARE SOMETIMES OVERRIDEN BUT NEED TO BE BLANK OTHERWISE -->
 		<platformSystemProperties></platformSystemProperties>
 		<moduleProperties></moduleProperties>


SimonCockx:
Hi @maoo! Nice work figuring that out. We're currently using Java 11, which is probably why it was working fine for us. I will test out your change in a bit, and double-check that it doesn't break something in my environment...

I hope you're all set-up. If you need any more help, feel free to contact us!

maoo:
> Hi @maoo! Nice work figuring that out. We're currently using Java 11, which is probably why it is working fine for us. I will test out your change in a bit, and double-check that it doesn't break something in my environment...

Thanks for the feedback! Not sure if I missed it from the docs, but couldn't find it written anywhere; it may be a good idea to share this info somewhere in the README; another good idea would be to use the [Maven Enforcer Plugin](https://maven.apache.org/enforcer/enforcer-rules/requireJavaVersion.html), so that the build will fail if Java version is different from 11.

As I started using Java 11, I decided to fallback to Tycho version `2.5.0`, as I saw more errors coming up ; seems that the Tycho version change needs a bit more testing.

> I hope you're all set-up. If you need any more help, feel free to contact us!

I am still facing issues (using Java 11), I shared the output on https://gist.github.com/maoo/b48b5e390d602ebab1fa74c2c42d80d2

Also tried adding `-Pmacos,jdk9-or-newer` to my `mvn` build command, but got the same error.

Thanks for the support!

maoo:
> @maoo The fact we use Java 11 is currently undocumented. I'll try to write a getting-started guide this (or next) week that should also have that info. Great idea to use the enforcer plugin for that too!
> 
> For the build issues you're facing: I haven't seen this one before. Is it possible that it is due to [eclipse-tycho/tycho#725](https://github.com/eclipse-tycho/tycho/issues/725)? Switching tycho versions can sometimes corrupt your `.m2` folder.

Nice catch @SimonCockx ! Another thing that could be mentioned in your "Getting Started" guide 😄 

Here's the command I'm using now - ` JAVA_HOME=`/usr/libexec/java_home -v11` mvn install -Pmacos,jdk9-or-newer -Dmaven.repo.local=/tmp/m2repo`

And this is the build error I'm getting now, which is different from the previous - https://gist.github.com/maoo/ce89f90d87b70511784d77d848f5a64b#file-build-output-txt-L1428

Is this still related with Java version? FYI, here my Maven setup info:

```
/private/tmp/rosetta-dsl(master*) » JAVA_HOME=`/usr/libexec/java_home -v11` mvn -v
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /usr/local/Cellar/maven/3.8.6/libexec
Java version: 11.0.16.1, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk-11.0.16.1.jdk/Contents/Home
Default locale: en_ES, platform encoding: UTF-8
OS name: "mac os x", version: "12.4", arch: "x86_64", family: "mac"
------------------------------------------------------------
```

SimonCockx:
> Nice catch @SimonCockx ! Another thing that could be mentioned in your "Getting Started" guide 😄
You're right. We haven't encountered this problem before, but it might help others that have been using a different Tycho version in the past. Thanks for being the frontier on opening this project up to a wider audience. :) 

> Here's the command I'm using now - ` JAVA_HOME=`/usr/libexec/java_home -v11` mvn install -Pmacos,jdk9-or-newer -Dmaven.repo.local=/tmp/m2repo`
> 
> And this is the build error I'm getting now, which is different from the previous - https://gist.github.com/maoo/ce89f90d87b70511784d77d848f5a64b#file-build-output-txt-L1428
> 
> Is this still related with Java version? FYI, here my Maven setup info:
Hm, strange. I'm especially bothered by [line 1432](https://gist.github.com/maoo/ce89f90d87b70511784d77d848f5a64b#file-build-output-txt-L1432): 
```
[ERROR] The method search(T, Function<T,Collection<T>>, Predicate<T>) from the type BreadthFirstSearch refers to the missing type Function
```
The type `Function` was added to Java 8, so it definitely shouldn't be missing. Your maven config seems okay. I'll try running with your config myself (except that I'm on Windows, but I don't think that is the issue here).

SimonCockx:
@maoo I fear I can't reproduce it. Could you maybe run a verbose `mvn clean install -X` (with your additional flags) and share your log?

SimonCockx:
Aha, I just had a discussion with a colleague and we were fearing you might run into this issue eventually. :) Two notes:
- Xsemantics is currently only released as a snapshot. I'm correct FINOS does not allow these kind of dependencies. One hour ago I've raised an issue to address this here: https://github.com/eclipse/xsemantics/issues/172
- We currently have added a repository to our `settings.xml` file to get Xsemantics working, but come to think of it, demanding that every developer adds it to their `settings.xml` doesn't sound very open-sourcy. We actually had a big discussion about this a while ago... Let me get back to this one tomorrow.

maoo:
> Aha, I just had a discussion with a colleague and we were fearing you might run into this issue eventually. :) Two notes:
> 
> * Xsemantics is currently only released as a snapshot. I'm correct FINOS does not allow these kind of dependencies. One hour ago I've raised an issue to address this here: [Xsemantics doesn't have a release build eclipse/xsemantics#172](https://github.com/eclipse/xsemantics/issues/172)

You can use `SNAPSHOT` dependencies in FINOS, as much as you want, but if you want to release under FINOS, you'll need to comply with Maven Central requirements, one of which is to depend only on released versions; a dependency against a SNAPSHOT version would break the release. 

> * We currently have added a repository to our `settings.xml` file to get Xsemantics working, but come to think of it, demanding that every developer adds it to their `settings.xml` doesn't sound very open-sourcy. We actually had a big discussion about this a while ago... Let me get back to this one tomorrow.

Could this be added to the main pom.xml, maybe in a profile that is activated by default (easy to remove if/when possible) ?

SimonCockx:
@maoo Good news: the maintainer of Xsemantics is very responsive and wants to do a release soon. Once that's done, Xsemantics can be fetched from Maven Central and we don't need to add any repositories to the `pom.xml`. (by the way: in March we actually tried adding the snapshot repo to the `pom.xml`, but apparently this is not added transitively to downstream projects, so we had to add it to every project of ours that depends on the Rosetta DSL, which is quite a lot) Anyway, it will all be setup properly soon. In the meantime, you can temporarily add this repo locally to get started:
```
<repository>
  <id>sonatype_snapshots</id>
  <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
  <releases>
    <enabled>false</enabled>
  </releases>
  <snapshots>
    <enabled>true</enabled>
  </snapshots>
</repository>
```

If you want, you can give Lorenzo a thumbsup for his release! https://github.com/eclipse/xsemantics/issues/172#issuecomment-1246382636

maoo:
Thanks @SimonCockx ! With the snapshot repos (and skipping tests) I managed to get a `BUILD SUCCESSFUL` from Maven, which is great. I've also added CVE scanning in the action, you can see the list of CVEs on https://github.com/maoo/rosetta-dsl/actions/runs/3056101256/jobs/4929907023#step:5:6260 (same list on https://github.com/REGnosys/rosetta-dsl/issues/401 ).

At this point, feel free to close this PR, since the proposed change is not really relevant any longer.

Thanks!

SimonCockx:
> Thanks @SimonCockx ! With the snapshot repos (and skipping tests) I managed to get a `BUILD SUCCESSFUL` from Maven, 

That's great news @maoo! Out of interest: did you encounter failing tests during the build? That shouldn't happen...

> At this point, feel free to close this PR, since the proposed change is not really relevant any longer.

Alright, yesterday I've been added to the FINOS Slack, so give me a shout there if anything else pops up. Glad I could help.

PR #410, opened 2022-09-29
Deprecation warnings for or-nodes and external functions

SimonCockx:


SimonCockx:
Replacing rule syntax with core syntax makes this PR obsolete.

PR #430, opened 2022-10-25
Story-1696  reports validation message

PayalKhanna:


com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/validation/ValidationResult.java
```diff
@@ -53,7 +53,7 @@ public ModelValidationResult(String name, ValidationType validationType, String
 			this.path = path;
 			this.modelObjectName = modelObjectName;
 			this.definition = definition;
-			this.failureReason = failureReason;
+			this.failureReason = getUpdatedFailureReason(name, modelObjectName, failureReason);
```
SimonCockx, line 56: I think it would be less surprising for developers using this class if the constructor was left as a simple record, i.e., for the following property to hold
```
var reason = Optional.of("The validation failed");
var validationResult = new ModelValidationResult(..., reason);
assertEquals(reason, validationResult.getFailureReason());
```
Instead, if we need the updated failure reason, we could just call `res.getUpdatedFailureReason()` and it would compute it on the fly, i.e., based on `this.name`, `this.modelObjectName` and `this.failureReason`.

On the other hand, I don't know where or how tightly coupled this class is to downstream projects, so I'll leave it up to you. What do you think?

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/validation/ValidationResult.java
```diff
@@ -82,6 +82,9 @@ public String getDefinition() {
 		
 		@Override
 		public Optional<String> getFailureReason() {
+			if (!failureReason.isEmpty() && modelObjectName.endsWith("Report")) {
+				return getUpdatedFailureReason();
+			}
```
SimonCockx, line 87: This change will again invalidate the expected property
```
var reason = Optional.of("The validation failed");
var validationResult = new ModelValidationResult(..., reason);
assertEquals(reason, validationResult.getFailureReason());
```
... but okay, I'll approve to get things rolling for you

SimonCockx (approved):

PR #457, opened 2022-12-01
STORY-1740: omit implicit parameter in operations and function calls

SimonCockx:
Features:
- Added synonym `it` for `item`. (need to decide on a final name for it!)
- Can omit the first parameter of _any_ operation in a context where an implicit parameter is present. (`extract [...]`, `sum`, `CallToFunc`, `+10`, etc, translates to `it extract [...]`, `it sum`, `CallToFunc(it)`, `it+10`)
- Conditions now have access to `item`/`it` to reference the instance being checked.
- Can directly access attributes of `item`/`it` wherever it exists, no need to write `item ->`.

Bug fixes (that I know of):
- addressed antlr warnings about ambiguous parsing paths.
- conditions are more robust. E.g., the following now runs fine.
```
type Foo
  a int (0..1)
  condition C:
    a exists
```
- nested inline functions are more robust.
- using `item` in a context where it doesn't exist is not allowed anymore.

and lots and lots of clean-ups.

Two particular challenging technicalities:
- Making the argument of a unary function call optional has as consequence that the parser cannot distinguish between variable references (`RosettaCallableCall`) and function calls (`RosettaCallableWithArgsCall`), because there might not be any parentheses. I therefore joint the concept into one: `RosettaSymbolReference`. This had a lot of impact in every part of the project.
- A question arises what to do with validation errors if the target of the error is an omitted parameter (e.g., omitting `it` in `it sum` where `it` is a list of `Foo` => should give a type error on `it`). Xtext by default just red-underlines the container (e.g., in this case `sum` would be red-underlined instead of `it`, which is okay for now), but Xsemantics filters out these validation errors. That needed a fix: see `RetainXsemanticsIssuesOnGeneratedInputsFilter`.


.gitignore
```diff
@@ -9,6 +9,7 @@ build.acceleo
 .metadata/
 *.class
 .DS_Store
+plugin.xml_gen
```
SimonCockx, line 12: This file is generated by xtext and shouldn't be in source control.

com.regnosys.rosetta.tests/src/com/regnosys/rosetta/generator/java/docrefs/DocReferenceTest.xtend
```diff
@@ -127,7 +127,7 @@ class DocReferenceTest {
 			func Sum:
 				[docReference Org1 Agr1 name "something"]
 				inputs: x number (0..*)
-				output: sum number (1..1)
+				output: s number (1..1)
```
SimonCockx, line 130: `sum` is a keyword and should not be used as a variable name. Otherwise there is an ambiguity in the parser. E.g., should `sum` be interpreted as the variable `sum` or as the operation `sum` applied to an implicit variable?

Alternative is to disallow applying `sum` to an implicit variable (e.g., you always need to specify the input: `it sum`)

com.regnosys.rosetta.tests/src/com/regnosys/rosetta/generator/java/expression/ListOperationTest.xtend
```diff
@@ -96,7 +98,7 @@ class ListOperationTest {
 						
 						protected List<Foo.FooBuilder> assignOutput(List<Foo.FooBuilder> filteredFoos, List<? extends Foo> foos) {
 							filteredFoos = toBuilder(MapperC.of(foos)
-								.filterItem(_item -> (Boolean)areEqual(_item.<Boolean>map("getInclude", __foo -> __foo.getInclude()), MapperS.of(Boolean.valueOf(true)), CardinalityOperator.All).get()).getMulti());
+								.filterItem(_«defaultImplicitVariable.name» -> (Boolean)areEqual(_«defaultImplicitVariable.name».<Boolean>map("getInclude", __foo -> __foo.getInclude()), MapperS.of(Boolean.valueOf(true)), CardinalityOperator.All).get()).getMulti());
```
SimonCockx, line 101: I've made tests robust against changes in the name of implicit variables (e.g., `item`, `it`, `this`, ...) It's now defined in a centralised class `ImplicitVariableUtil`.

com.regnosys.rosetta.tests/src/com/regnosys/rosetta/generator/java/function/FunctionGeneratorTest.xtend
```diff
@@ -35,6 +35,55 @@ class FunctionGeneratorTest {
...
+			
+			func F1:
+				inputs:
+					foos Foo (0..*)
+				output:
+					result int (0..*)
+				
+				add result:
+					foos
+						extract [ a + b count]
```
SimonCockx, line 56: We now never need to write `item -> ...`. You can just leave the `item` (or `it`, or whatever you want to call it) Hooray!

com.regnosys.rosetta.tests/src/com/regnosys/rosetta/generator/java/function/FunctionGeneratorTest.xtend
```diff
@@ -35,6 +35,55 @@ class FunctionGeneratorTest {
...
+			version "${project.version}"
+			
+			func F1:
+				inputs:
+					a int (0..*)
+				output:
+					result int (0..*)
+				
+				add result:
+					a extract [+ 1]
```
SimonCockx, line 79: The left argument of _every_ operation (`extract`, `sum`, and yes, even `+`) is now implicitly filled with `item` when left out.

com.regnosys.rosetta.tests/src/com/regnosys/rosetta/generator/java/function/FunctionGeneratorTest.xtend
```diff
@@ -3571,7 +3620,11 @@ class FunctionGeneratorTest {
 							return ValidationResult.success(NAME, ValidationResult.ValidationType.DATA_RULE,  "Foo", path, DEFINITION);
 						}
 						
-						return ValidationResult.failure(NAME, ValidationResult.ValidationType.DATA_RULE, "Foo", path, DEFINITION, result.getError());
+						String failureMessage = result.getError();
+						if (failureMessage == null) {
+							failureMessage = "Condition " + NAME + " failed.";
+						}
+						return ValidationResult.failure(NAME, ValidationResult.ValidationType.DATA_RULE, "Foo", path, DEFINITION, failureMessage);
```
SimonCockx, line 3627: See changes to `DataRuleGenerator`. Made this more robust, as conditions could fail in trivial cases like `condition C: a exists` where `a` is of type `int (0..1)`.

com.regnosys.rosetta.tests/src/com/regnosys/rosetta/generator/java/rule/ConditionTestHelper.xtend
```diff
@@ -8,20 +8,47 @@ import com.rosetta.model.lib.path.RosettaPath
...
+import com.rosetta.model.lib.functions.ModelObjectValidator
+import com.rosetta.model.lib.functions.NoOpModelObjectValidator
 
 class ConditionTestHelper {
 	
 	@Inject extension ModelHelper
 	@Inject extension RosettaExtensions
 	@Inject extension CodeGeneratorTestHelper
 	
+	final Injector injector
```
SimonCockx, line 25: Currently, the test code responsible for running conditions didn't inject its dependencies like functions that the condition needs to call. Fixed here.

com.regnosys.rosetta.tests/src/com/regnosys/rosetta/tests/RosettaExpressionsTest.xtend
```diff
@@ -48,9 +48,11 @@ class RosettaExpressionsTest {
 		model => [
 			((elements.last as Function)
 			  .operations.head.expression as RosettaConditionalExpression) => [
-			  	isFull.assertFalse;
-			  	elsethen.assertNotNull;
-			  	eqHelper.equals(elsethen, ExpressionFactory.eINSTANCE.createListLiteral).assertTrue
+			  	assertFalse(isFull);
+			  	assertNotNull(elsethen);
+			  	val lit = ExpressionFactory.eINSTANCE.createListLiteral
+			  	lit.generated = true
```
SimonCockx, line 54: I added a property `generated` to `RosettaExpression` to see whether the expression was generated by the `RosettaDerivedStateComputer` as syntactic sugar or not.

com.regnosys.rosetta/model/Rosetta.xcore
```diff
@@ -307,30 +312,59 @@ class RosettaSegment extends RosettaRootElement, RosettaNamed {
  * Mappings
  */
 
+interface RosettaMapTest {}
+
+interface RosettaMapTestExpression extends RosettaMapTest {}
```
SimonCockx, line 317: I separated the rosetta map test expression syntax from the rosetta expression syntax.

com.regnosys.rosetta/model/RosettaExpression.xcore
```diff
@@ -52,17 +66,50 @@ class RosettaIntLiteral extends RosettaLiteral {
 	}
 }
 
-class ListLiteral extends RosettaLiteral {
+/*
+ * List construction
+ */
+
+class ListLiteral extends RosettaExpression {
```
SimonCockx, line 73: Throughout the code there is a confusion between "literals" and "constants"... A `ListLiteral` is definitely not constant, so I removed it from the `RosettaLiteral` hierarchy. Perhaps we should rename `RosettaLiteral` to `RosettaConstant`? Not sure about the terminology.

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -410,9 +418,9 @@ RosettaSegment:
  * Expressions 
  */ 
 
-RosettaCallableCall: 
-	callable=[RosettaCallable|QualifiedName]
-	| implicitReceiver ?= 'item'
+RosettaReferenceOrFunctionCall returns RosettaExpression:
+	  {RosettaSymbolReference} symbol=[RosettaSymbol|QualifiedName] (explicitArguments?='(' (args+=RosettaCalcExpression (',' args+=RosettaCalcExpression)*)? ')')?
+	| {RosettaImplicitVariable} name=('item' | 'it')
```
SimonCockx, line 423: I added the option to use `it` instead of `item`... We should decide on a final name though.

com.regnosys.rosetta/src/com/regnosys/rosetta/generator/java/util/JavaNames.xtend
```diff
@@ -81,8 +81,6 @@ class JavaNames {
 			RosettaRecordType:
 				JavaType.create(JavaClassTranslator.toJavaFullType(type.name)) ?:
 					JavaType.create(packages.defaultLibRecords.name + '.' + type.name.toFirstUpper)
-			RosettaExternalFunction:
-				createJavaType(packages.defaultLibFunctions, type.name)
```
SimonCockx, line 85: A `RosettaExternalFunction` used to be a `RosettaType` for some reason. I fixed that.

com.regnosys.rosetta/model/RosettaExpression.xcore
hugohills-regnosys: RosettaPrimitive perhaps?

hugohills-regnosys (approved):

PR #458, opened 2022-12-02
STORY-459: A modeller should be able to consistently use a colon for both functions and reporting rules

SimonCockx:


com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -678,8 +678,9 @@ RosettaRuleReference:
 	'[' 'ruleReference' reportingRule=[RosettaBlueprint] ']'	
 ;
 
+// @Compat: make ':' mandatory at January 16
```
SimonCockx, line 681: I'm thinking of making a comment such as this one for any feature that needs transitioning, so we don't forget about it.

hugohills-regnosys: Yeah, it's a good idea

hugohills-regnosys (approved):

PR #478, opened 2022-12-20
Added support to extend multiple synonyms

minesh-s-patel:


com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -681,7 +681,7 @@ RosettaOnlyExistsElementRoot returns RosettaSymbolReference:
  *****************************************/
 
 RosettaExternalSynonymSource:
-	'synonym' 'source' RosettaNamed ('extends' superSynonym=[RosettaSynonymSource]) '{'
+		'synonym' 'source' RosettaNamed ('extends' (superSynonyms += [RosettaSynonymSource] (',' superSynonyms += [RosettaSynonymSource] )*)? )? '{'
```
SimonCockx, line 684: I think there is one question mark too much here. Now you allow syntax such as

```
synonym source Foo extends {

}
```
I.e., it should be
```
'synonym' 'source' RosettaNamed ('extends' superSynonyms += [RosettaSynonymSource] (',' superSynonyms += [RosettaSynonymSource])* )? '{'
```

SimonCockx (approved):

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
hugohills-regnosys: Good catch

PR #480, opened 2023-01-03
Add Java 8 support

hugohills-regnosys:


pom.xml
```diff
@@ -35,8 +35,14 @@ See the Apache License Version 2.0 for the specific language governing permissio
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<enforcer.version>3.1.0</enforcer.version>
 		<java.enforced.version>[11,12)</java.enforced.version>
+		<!-- source and target used by tycho plugin -->
 		<maven.compiler.source>11</maven.compiler.source>
 		<maven.compiler.target>11</maven.compiler.target>
```
SimonCockx, line 40: Might as well get rid of `source` and `target`, as it is pretty confusing that they're here while not being used.

pom.xml
```diff
@@ -35,8 +35,14 @@ See the Apache License Version 2.0 for the specific language governing permissio
...
+		<!-- used by maven compiler plugin -->
+		<maven.compiler.release>11</maven.compiler.release>
+		<!-- for compatibility, release version is java 8 in lib projects -->
+		<lib.maven.compiler.target>1.8</lib.maven.compiler.target>
```
SimonCockx, line 44: Same here

com.regnosys.rosetta.lib/pom.xml
```diff
@@ -11,16 +11,25 @@
...
+	</properties>
+
     <build>
         <plugins>
             <plugin>
                 <groupId>org.eclipse.xtend</groupId>
                 <artifactId>xtend-maven-plugin</artifactId>
+                <configuration>
+                	<javaSourceVersion>${maven.compiler.release}</javaSourceVersion>
+                </configuration>
```
SimonCockx, line 27: Actually, I don't think we need this plugin at all in `lib`. We don't (and shouldn't) use xtend in our lib project.

SimonCockx (approved):

com.regnosys.rosetta.lib/pom.xml
hugohills-regnosys: I agree we shouldn't, but we do at the moment `com.rosetta.model.lib.expression.MapperMaths.xtend`

PR #490, opened 2023-01-18
STORY-468: Expression formatting

SimonCockx:


com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/annotations/RosettaChoiceRule.java
```diff
@@ -4,6 +4,8 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
+// @Compat. This is not used anymore. All conditions are now `RosettaDataRule`s.
+@Deprecated
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
 public @interface RosettaChoiceRule {
```
SimonCockx, line 11: Cleanup 1: there is only one type of condition; `RosettaDataRule`s. `RosettaChoiceRule`s are deprecated.

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/meta/RosettaMetaData.java
```diff
@@ -15,7 +16,11 @@
 
 	List<Validator<? super T>> dataRules(ValidatorFactory factory);
 	
-	List<Validator<? super T>> choiceRuleValidators();
+	// @Compat. This will be empty, as choice rules are now all data rules.
+	@Deprecated
+	default List<Validator<? super T>> choiceRuleValidators() {
+		return Collections.emptyList();
+	}
```
SimonCockx, line 23: Part of cleanup 1: no need to implement `choiceRuleValidators` anymore. Just return an empty list by default.

com.regnosys.rosetta/model/Rosetta.xcore
```diff
@@ -287,20 +287,30 @@ enum RosettaSynonymRef {
...
+class RosettaDocReference {
+	contains RegulatoryDocumentReference docReference
+	contains DocumentRationale[] rationales
+	String structuredProvision
 	String provision
-	contains RosettaSegmentRef[] segments
 	boolean reportedField
+	// @Compat. In DRR, there is a segment `table` located after a `rationale`. This should never be the case, but to remain backwards compatible, we need to allow this.
+	contains RosettaSegmentRef[] extraneousSegments
+}
```
SimonCockx, line 298: Cleanup 2: added `rationale`, `rationale_author` and `structured_provision` as actual keywords rather than document segments.

com.regnosys.rosetta/model/RosettaExpression.xcore
```diff
@@ -271,6 +272,21 @@ class LastOperation extends ListOperation {
...
+
+enum Necessity {
+	optional, required
+}
+
+class ChoiceOperation extends RosettaUnaryOperation {
+	Necessity necessity
+	refers Attribute[] attributes
+}
+
```
SimonCockx, line 289: Part of cleanup 1: `one-of` and `choice` are now actual operations.

com.regnosys.rosetta/model/RosettaSimple.xcore
```diff
@@ -90,20 +90,9 @@ class FunctionDispatch extends Function {
...
-class Constraint {
-	boolean oneOf
-	Necessity necessity
-	refers Attribute[] attributes
-}
-
-enum Necessity {
-	optional, required
-}
-
```
SimonCockx, line 106: Part of cleanup 1: constraints don't exist anymore.

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -112,24 +115,13 @@ ShortcutDeclaration:
 Condition:
 	'condition' RosettaNamed? ':' =>Definable?
 		=>(References|Annotations)* // conflicts with ListLiteral as RosettaCalcExpression
-		( constraint = OneOf
-		| constraint = Constraint
-		| expression = RosettaCalcExpression)
```
SimonCockx, line 117: Cleanup 1: `choice` conditions don't exist anymore. There are only "data rules".

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -142,19 +134,19 @@ Operation:
 // @Compat deprecated - to be deleted once all models updated
 AssignOutputOperation:
 	'assign-output' assignRoot = [AssignPathRoot|ValidID] (path = Segment)? ':' =>Definable?
-	expression = RosettaCalcExpression assignAsKey?='as-key'?
```
SimonCockx, line 145: Cleanup 2: `as-key` is now a special kind of operation instead of a keyword.

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -142,19 +134,19 @@ Operation:
...
 
 OutputOperation:
 	('set' | add ?= 'add') assignRoot = [AssignPathRoot|ValidID] (path = Segment)? ':' =>Definable?
-		expression = RosettaCalcExpression assignAsKey?='as-key'?
+		expression = RosettaCalcExpressionWithAsKey
 ;
 
 Segment:
-	FEATURE_CALL_SEPARATOR attribute = [Attribute|ValidID] (next = Segment)?;
+	'->' attribute = [Attribute|ValidID] (next = Segment)?;
```
SimonCockx, line 146: Cleanup 3: I made `->` a keyword instead of a terminal rule. Makes it much easier to format.

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -405,23 +397,32 @@ RosettaMapPathValue:
...
 RosettaDocReference:
-	'[' ('regulatoryReference'|'docReference') body=[RosettaBody|ValidID] corpuses+=([RosettaCorpus|ValidID])+ (segments+=RosettaSegmentRef)+ ('provision' provision=STRING)? (reportedField?='reportedField')?  ']'
+	'[' ('regulatoryReference'|'docReference') docReference=RegulatoryDocumentReference
+		(rationales+=DocumentRationale)*
+		// @Compat. In DRR, there is a segment `table` located after a `rationale`. This should never be the case, but to remain backwards compatible, we need to allow this.
+ 		// NOTE: once we get rid of `extraneousSegments`, we can also remove the syntactic predicate `=>` on the segment of rule `RegulatoryDocumentReference`.
+ 		(extraneousSegments+=RosettaSegmentRef)*
+		('structured_provision' structuredProvision=STRING)?
+		('provision' provision=STRING)?
+		(reportedField?='reportedField')? ']'
```
SimonCockx, line 407: Cleanup 4: `rationale`, `rationale_author` and `structured_provision` are now actual keywords instead of document segments.

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -765,19 +773,18 @@ RosettaBlueprint:
 ;
 
 BlueprintNodeExp:
-	node=BlueprintNode (reference = RosettaDocReference)? ('then' next=BlueprintNodeExp)? 
+	node=BlueprintNode (reference=RosettaDocReference)?
+	('then' next=BlueprintNodeExp)?
+	(=>'as' identifier=STRING)?
```
SimonCockx, line 778: Cleanup 5: moved `as` up one level in the AST. Makes it easier to format.

com.regnosys.rosetta.lib/META-INF/MANIFEST.MF
```diff
@@ -18,7 +18,8 @@ Require-Bundle: org.eclipse.xtext,
  com.google.guava,
  org.eclipse.xtext.xbase.lib,
  org.eclipse.xtend.lib,
- org.eclipse.xtend.lib.macro
+ org.eclipse.xtend.lib.macro,
+ org.apache.commons.lang3
```
SimonCockx, line 22: Added this dependency. Should be okay, I think.

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/annotations/RosettaChoiceRule.java
SimonCockx, line 11: I made `RosettaChoiceRule`. This doesn't modify any behaviour, so 100% backwards compatible. The code generator does not use this annotation anymore.

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
```diff
@@ -262,4 +268,30 @@ private static <T> String formatMultiError(Mapper<T> o) {
 				t.getClass().getSimpleName() : // for rosettaModelObjects only log class name otherwise error messages are way too long
 				o.getMulti().toString();
 	}
+	
+	// one-of and choice
+
+	public static <T> ComparisonResult choice(Mapper<T> mapper, List<String> choiceFieldNames, ValidationResult.ChoiceRuleValidationMethod necessity) {
```
SimonCockx, line 274: I added a single static method to `ExpressionOperators`. Didn't remove anything, so 100% backwards compatible.

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/meta/RosettaMetaData.java
SimonCockx, line 23: The code generator will not implement this method anymore. By providing a default implementation, this is okay.

com.regnosys.rosetta.ide/src/com/regnosys/rosetta/ide/formatting/RosettaFormattingService.java
```diff
@@ -0,0 +1,69 @@
...
+			if (options.isInsertSpaces()) {
+				indent = Strings.padEnd("", options.getTabSize(), ' ');
+			}
+		}
+		List<TextEdit> result = new ArrayList<>();
+		if (this.formatter2Provider != null) {
+			MapBasedPreferenceValues preferences = new MapBasedPreferenceValues();
+			preferences.put("indentation", indent);
+			
+			Number conditionalMaxLineWidth = options.getNumber("conditionalMaxLineWidth");
```
minesh-s-patel, line 47: We should use constants for these. I saw some tests referring to these string.

com.regnosys.rosetta.ide/src/com/regnosys/rosetta/ide/formatting/RosettaFormattingService.java
```diff
@@ -0,0 +1,69 @@
...
+			FormattingOptions options) {
+		String indent = indentationInformation.getIndentString();
+		if (options != null) {
+			if (options.isInsertSpaces()) {
+				indent = Strings.padEnd("", options.getTabSize(), ' ');
+			}
+		}
+		List<TextEdit> result = new ArrayList<>();
+		if (this.formatter2Provider != null) {
+			MapBasedPreferenceValues preferences = new MapBasedPreferenceValues();
```
minesh-s-patel, line 44: Perhaps move the `preferences` into its own method to improve readability

com.regnosys.rosetta.ide/src/com/regnosys/rosetta/ide/highlight/RosettaHighlightingCalculator.xtend
```diff
@@ -25,7 +25,7 @@ import org.eclipse.xtext.nodemodel.util.NodeModelUtils
 import org.eclipse.xtext.util.CancelIndicator
 
 import static com.regnosys.rosetta.rosetta.RosettaPackage.Literals.*
-import com.regnosys.rosetta.rosetta.RosettaDocReference
+import com.regnosys.rosetta.rosetta.RegulatoryDocumentReference
```
minesh-s-patel, line 28: This is a non-backward compatible change. This means that this release is blocked from being used in Rosetta. All ecore changes as this point must be backward compatible.

com.regnosys.rosetta.ide/src/com/regnosys/rosetta/ide/server/RosettaServerModule.java
```diff
@@ -12,10 +15,20 @@ public class RosettaServerModule extends AbstractGenericModule {
...
 	 * Do not use the constructor. Use {@code RosettaServerModule.create()} instead.
 	 */
-	private RosettaServerModule() {}
+	protected RosettaServerModule() {}
 	
 	public static Module create() {
-		return Modules.override(new ServerModule()).with(new RosettaServerModule());
+		return create(RosettaServerModule.class);
+	}
+	public static Module create(Class<? extends RosettaServerModule> serverModuleClass) {
```
minesh-s-patel, line 23: Why do this reflection here?

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
minesh-s-patel: Not sure I understand why we need to reflectively create this type.
If we need it for backward compatibility, then we should refactor it out into `BackwardCompatibilityChoice` type.

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
```diff
@@ -262,4 +268,30 @@ private static <T> String formatMultiError(Mapper<T> o) {
...
+		T object = mapper.get();
+		List<String> populatedFieldNames = new LinkedList<>();
+		for (String a: choiceFieldNames) {
+			try {
+				Method getter = object.getClass().getMethod("get" + StringUtils.capitalize(a));
+				if (ExistenceChecker.isSet(getter.invoke(object))) {
+					populatedFieldNames.add(a);
+				}
+			} catch (NoSuchMethodException | SecurityException | IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+				throw new IllegalArgumentException(e);
```
minesh-s-patel, line 284: What is the impact of throwing here? Does the code gen need to handle `IllegalArgumentException`?

com.regnosys.rosetta/model/Rosetta.xcore
```diff
@@ -287,20 +287,30 @@ enum RosettaSynonymRef {
  * docReferences
  */
 
-class RosettaDocReference extends RosettaRegulatoryBody {
+class RosettaDocReference {
```
minesh-s-patel, line 290: Sneaky :-)

minesh-s-patel: You think we should deprecate this type? Usages should be migrated to `RegulatoryDocumentReference`

com.regnosys.rosetta/src/com/regnosys/rosetta/Rosetta.xtext
minesh-s-patel: This actually makes the `as` a lot more useful in DRR. Even though we want to eventually deprecate and remove it.

com.regnosys.rosetta.ide/src/com/regnosys/rosetta/ide/formatting/RosettaFormattingService.java
SimonCockx: Agreed. Most of this code is copied over from the parent class (including the magic string `"indentation"`), I've only made a couple of adjustements. But yeah, it's not because Xtext doesn't do it the right way, that we should, so I'll adjust that.

com.regnosys.rosetta.ide/src/com/regnosys/rosetta/ide/highlight/RosettaHighlightingCalculator.xtend
minesh-s-patel: I take it back. I can see its backwards compatible further in the review.

minesh-s-patel (approved):

com.regnosys.rosetta.ide/src/com/regnosys/rosetta/ide/highlight/RosettaHighlightingCalculator.xtend
SimonCockx: I don't think this kind of change will cause a problem. If I push this through, Rosetta will be able to handle this ecore change.

Or is this related to the BSP work?

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
SimonCockx: This won't throw during code gen, but it can throw when using generated code.

However, it'll only throw if you actually provide a nonsense argument. For example, if you have a type with attributes called `foo` and `bar`, then you should only pass in these attribute names. If you pass in `party`, this will throw.

This will of course never happen in the code we generate. It's only when a consumer of our generated code tries to do something wrong that this will throw.

com.regnosys.rosetta.lib/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
SimonCockx: This one is not for backwards compatibility. It's the runtime support for the "new" operations `one-of` and `optional/required choice`.

An expression `optional choice foo, bar`, which is syntactic sugar for `item optional choice foo, bar`, will generate more or less the following code.
```
choice(item, Lists.of("foo", "bar"), ChoiceRuleValidationMethod.OPTIONAL)
```

There are a couple of alternatives to using reflection here, but I'm not sure what the best option would be:
1. instead of representing the attributes as `String`s, we could represent them as lambdas, i.e., `Lists.of(a -> a.getFoo(), a -> a.getBar())`. This is definitely possible, but a bit more tricky to get right.
2. we could generate a runtime representation of attributes in every `RosettaModelObject` class. Would require a bit more work.
3. we could implement it in the way that `only exists` works by passing in the evaluated attributes directly. However, the way that it works in `only exists` is terrible: it will evaluate parent object again for each attribute. For `only exists` this is actually okay because parent expressions are restricted to simple feature calls. That's not the case here though.

So... I think (1) is better, (2) is the best but more work, and (3) is terrible. How I've done it now is just the simple way.

com.regnosys.rosetta/model/Rosetta.xcore
SimonCockx: I think the naming is confusing, because they are both actually needed. :) A `RosettaDocReference` represents the annotation, while `RegulatoryDocumentReference` represents a reference to a part of a regulatory document, e.g., a body, a list of corpuses and a list of segments. `RosettaDocReference` contains a `RegulatoryDocumentReference`, but also `rationale`s, `provision`, etc.

PR #506, opened 2023-03-10
STORY-509: A user can leave out brackets and use the `then` keyword to chain operations together.

SimonCockx:


rosetta-ide/pom.xml
```diff
@@ -45,6 +45,15 @@
...
 
     <build>
+        <resources>
+            <resource>
+                <directory>${project.basedir}/src-gen/main/java</directory>
+                <includes>
+                    <include>META-INF/services/org.eclipse.xtext.ISetup</include>
+                </includes>
+            </resource>
+        </resources>
```
SimonCockx, line 55: Fix for issue David and James were experiencing on the BSP since the Tycho removal.

PR #596, opened 2023-07-14
Stale validator

hugohills-regnosys:

## Type of change


rosetta-lang/src/main/java/com/regnosys/rosetta/generator/java/object/ValidatorsGenerator.xtend
```diff
@@ -149,11 +151,12 @@ class ValidatorsGenerator {
 		if (attr.inf === 0 && attr.isUnbound) {
 			null
 		} else {
+	        /* Casting is required to ensure types are output to ensure code generation in Rosetta */
```
SimonCockx, line 154: This is confusing. Should be "to ensure _recompilation_ in Rosetta", I think.

hugohills-regnosys: You're right, we can update the comment

PR #637, opened 2023-09-06
JSON-like constructors

SimonCockx:
See [issue](https://github.com/REGnosys/rosetta-dsl/issues/398) for details.

## Type of change


rosetta-ide/.classpath
```diff
@@ -62,5 +62,11 @@
 			<attribute name="test" value="true"/>
 		</attributes>
 	</classpathentry>
+	<classpathentry excluding="**" kind="src" output="target/test-classes" path="src/test/resources">
```
hugohills-regnosys, line 65: Is this change expected?

rosetta-runtime/.classpath
```diff
@@ -67,5 +67,11 @@
 			<attribute name="test" value="true"/>
 		</attributes>
 	</classpathentry>
+	<classpathentry excluding="**" kind="src" output="target/test-classes" path="src/test/resources">
```
hugohills-regnosys, line 70: Is this expected

documentation/rosetta-modelling-component.md
```diff
@@ -662,6 +663,66 @@ E.g.:
...
+  favouriteAnimal: "bear",
+  mentor: Employee {
+    name: "Michael Scott",
+    age: empty,
+    favouriteAnimal: empty,
+    mentor: empty
+  }
+}
+```
+
```
lolabeis, line 699: I would add the alternative way of creating the "Dwight Schrute" record, which is first to set "Michael Scott" as an alias, and then use that alias when constructing Dwight.

I also understand that you can use Rosetta expressions within the instantiation syntax, for instance a path extraction, so you could add an example accordingly.

Those 2 points are described in the issue.

documentation/rosetta-modelling-component.md
```diff
@@ -662,6 +663,66 @@ E.g.:
...
+```
+<type name> {
+  <field name 1>: <field value 1>,
+  <field name 2>: <field value 2>,
+  etc
+}
+```
+
+As an example, consider the following data type representing an employee of a company.
+``` Haskell
```
lolabeis, line 678: Our "theme" for examples throughout the documentation is related to vehicles. I think we should continue to follow the same convention. And accordingly, those examples could be added to the demo model.

documentation/rosetta-modelling-component.md
```diff
@@ -662,6 +663,66 @@ E.g.:
...
+    age: empty,
+    favouriteAnimal: empty,
+    mentor: empty
+  }
+}
+```
+
+#### Triple-Dot Syntax
+
+Notice that in the example above most fields of Michael Scott are actually `empty`. Types with many optional attributes is common practice in Rosetta models such as the CDM, and assigning `empty` to them explicitly can be verbose. We can solve this by using the triple-dot keyword `...`, which will implicitly assign `empty` to all absent attributes.
```
lolabeis, line 702: Typo in the second sentence, not syntactically correct. "Having" missing perhaps?

davidalk (approved):

PR #666, opened 2023-10-06
Allow test on multiple platforms

natrem:
When running tests on a windows platform, files are created with local platform end of line character. On the other hand, the expected result of the test are static files (arbitrarily) containing *LF* end of line.

## Possible approaches
I chose to change the end of line characters in the expected output for the test, replacing them by the *platform* one.

Other possibilities include:
- change the actual generation to generate with a common end of line -> this would change the actual behavior of rosetta (although in a minor way). Generated files on windows platform would not match what the user would expect so I rejected that
- change the *actual* result (first commit) in the test by replacing the *platform* end of line by the *arbitrary* one. It seems less coherent to change that.

## Type of change


SimonCockx:
Hey @natrem, hmm, strange. We have been building this project on Mac, Linux and Windows just fine (in fact, I'm running the build on Windows locally all the time). So I'm confused to see these tests failing for you. Perhaps you have a different git configuration for line endings?

For me, it's
```
> git config core.autocrlf
true
```

PR #671, opened 2023-10-18
Improved stability of language server

SimonCockx:
ISSUE-861: Long if-then-else statement crashes BSP due to a null pointer exception in the ANTLR parser

Improvements:
- Unexpected errors during parsing (e.g., a `StackOverflowException`) are now handled gracefully (catch, log, return diagnostic to the user)
- Completion requests to the language server are now cancellable, which means they don't block subsequent requests if they run for a long time.
- Long requests to the language server will now time out after a configurable amount of seconds (configurable through an environment variable)

## Type of change

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@
 		<maven-clean-plugin.version>3.3.1</maven-clean-plugin.version>
 		<build-helper-maven-plugin.version>3.4.0</build-helper-maven-plugin.version>
 		<maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
+		<maven-source-plugin.version>3.3.0</maven-source-plugin.version>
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -394,6 +395,18 @@
...
+				<artifactId>maven-source-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>attach-sources</id>
+						<goals>
+							<goal>jar</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
```
SimonCockx, line 409: I noticed we weren't generating sources since June, although they make it easier to debug in downstream projects. I'm not sure why I have to include this plugin explicitly though, since I think other projects do this by default.

hugohills-regnosys (approved):

PR #679, opened 2023-10-25
Story-653 - Rewording Error messages to be more efficient when looking at remediation.

PayalKhanna:
Rewording Error message displayed on diagnostic panels to be more efficient when looking at remediation.

## Type of change


rosetta-runtime/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
```diff
@@ -46,7 +46,11 @@ public static <T> ComparisonResult exists(Mapper<T> o) {
 		if (o.resultCount()>0) {
 			return ComparisonResult.success();
 		}
-		return ComparisonResult.failure(o.getErrorPaths() + " does not exist");
+		String failuremessage = null;
+		if (o.getErrorPaths()!= null){
```
hugohills-regnosys, line 50: Is not null or empty

hugohills-regnosys (approved):

PR #683, opened 2023-11-02
Story 658 - sending Condition name in error message for all DataRule validationType

PayalKhanna:
Updated the Condition to be refactor the error message to include condition name for all Data Rule validation. 
checks for Null and contains text null to just send Condition has failed for error message which is comparing against null value and will display weird error message which are difficult to be interpreted by Users

## Type of change

### DIFF
--- a/rosetta-runtime/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
+++ b/rosetta-runtime/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
@@ -46,11 +46,7 @@ public class ExpressionOperators {
 		if (o.resultCount()>0) {
 			return ComparisonResult.success();
 		}
-		String failureMessage = null;
-		if (o.getErrorPaths()!= null && !o.getErrorPaths().isEmpty() && !o.getErrorPaths().toString().contains("Null")){
-			failureMessage = o.getErrorPaths() + " does not exist";
-		}
-		return ComparisonResult.failure(failureMessage);
+		return ComparisonResult.failure(o.getErrorPaths() + " does not exist");
 	}
 	
 	// singleExists
--- a/rosetta-runtime/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
+++ b/rosetta-runtime/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
@@ -47,7 +47,7 @@ public class ExpressionOperators {
 			return ComparisonResult.success();
 		}
 		String failureMessage = null;
-		if (o.getErrorPaths()!= null && !o.getErrorPaths().isEmpty()){
+		if (o.getErrorPaths()!= null && !o.getErrorPaths().isEmpty() && !o.getErrorPaths().toString().contains("Null")){
 			failureMessage = o.getErrorPaths() + " does not exist";
 		}
 		return ComparisonResult.failure(failureMessage);
--- a/rosetta-testing/src/test/java/com/regnosys/rosetta/generator/java/condition/DataRuleGeneratorTest.xtend
+++ b/rosetta-testing/src/test/java/com/regnosys/rosetta/generator/java/condition/DataRuleGeneratorTest.xtend
@@ -82,7 +82,7 @@ class DataRuleGeneratorTest {
 							
 							String failureMessage = result.getError();
 							if (failureMessage == null) {
-								failureMessage = "Condition " + NAME + " failed.";
+								failureMessage = "Condition has failed.";
 							}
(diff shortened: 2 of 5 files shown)


rosetta-runtime/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
```diff
@@ -47,7 +47,7 @@ public static <T> ComparisonResult exists(Mapper<T> o) {
 			return ComparisonResult.success();
 		}
 		String failureMessage = null;
-		if (o.getErrorPaths()!= null && !o.getErrorPaths().isEmpty()){
+		if (o.getErrorPaths()!= null && !o.getErrorPaths().isEmpty() && !o.getErrorPaths().toString().contains("Null")){
```
SimonCockx, line 50: Q: why are some error paths null? This seems a bit of a hack to workaround a bug. It is probably better to find out why we are generating error paths containing a string "Null".

rosetta-testing/src/test/java/com/regnosys/rosetta/generator/java/condition/DataRuleGeneratorTest.xtend
```diff
@@ -81,8 +81,8 @@ class DataRuleGeneratorTest {
 							}
 							
 							String failureMessage = result.getError();
-							if (failureMessage == null) {
-								failureMessage = "Condition " + NAME + " failed.";
+							if (failureMessage == null || failureMessage.contains("Null") || failureMessage == "") {
```
SimonCockx, line 84: Same here.

rosetta-testing/src/test/java/com/regnosys/rosetta/generator/java/condition/DataRuleGeneratorTest.xtend
```diff
@@ -198,8 +198,8 @@ class DataRuleGeneratorTest {
...
```
SimonCockx, line 201: And here.

rosetta-testing/src/test/java/com/regnosys/rosetta/generator/java/function/FunctionGeneratorTest.xtend
```diff
@@ -4510,8 +4510,8 @@ class FunctionGeneratorTest {
 							}
 							
 							String failureMessage = result.getError();
-							if (failureMessage == null) {
-								failureMessage = "Condition " + NAME + " failed.";
+							if (failureMessage == null || failureMessage.contains("Null") || failureMessage == "") {
```
SimonCockx, line 4513: And here.

rosetta-runtime/src/main/java/com/rosetta/model/lib/expression/ExpressionOperators.java
PayalKhanna: these are coming from when condition is within a function which is passed a  value like x-> y-> z-> and it return the last object as "null" as it doesn't know which object was actually passed.

PayalKhanna: o.getErrorPaths()!= null && !o.getErrorPaths().isEmpty() this did not return true and its was not empty or null but rather it had value "Null"

rosetta-testing/src/test/java/com/regnosys/rosetta/generator/java/condition/DataRuleGeneratorTest.xtend
PayalKhanna: this is to factor in the Null operand error message to  just return `Condition has failed` instead of wired messages

PayalKhanna: `EMIR_VR_2023_03:- Null operand null ExecutionTimestamp 2018-04-01T14 15 36Z`

PayalKhanna: `EMIR_VR_2060_01:- Null operand null Integer 0 and AssetClass INTR does not equal String EQUI and AssetClass INTR does not equal String COMM
Null operand: [[] : null] >= [[Integer] : 0]`

PayalKhanna: for error message like this it would return `EMIR_VR_2060_01:- Condition has failed`

SimonCockx (approved):

PR #687, opened 2023-11-13
Story 674 - task 3660 Fix RequestManager memory leak

davidalk:
This change fixes a leak in the XText RequestManager where the request store grows continuously when on read request are sent

## Type of change

### DIFF
--- a/rosetta-ide/src/main/java/com/regnosys/rosetta/ide/server/RosettaRequestManager.java
+++ b/rosetta-ide/src/main/java/com/regnosys/rosetta/ide/server/RosettaRequestManager.java
@@ -1,5 +1,6 @@
 package com.regnosys.rosetta.ide.server;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.eclipse.xtext.ide.server.concurrent.AbstractRequest;
 import org.eclipse.xtext.ide.server.concurrent.RequestManager;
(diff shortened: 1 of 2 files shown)


hugohills-regnosys (approved):

rosetta-ide/src/main/java/com/regnosys/rosetta/ide/server/RosettaRequestManager.java
```diff
@@ -35,7 +32,14 @@ public class RosettaRequestManager extends RequestManager {
 	                        .setDaemon(true)
 	                        .setNameFormat("rosetta-language-server-request-timeout-%d")
 	                        .build());
-		
+
+	/*
+	 * The code that uses this list fixes a memory leak in the RequestManager and should be contributed
+	 * back to the Xtext project then removed from here
+	 */
+	@VisibleForTesting
```
SimonCockx, line 40: The [docs for `VisibleForTesting`](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/annotations/VisibleForTesting.html) say the following:

> *Do not use this interface* for public or protected declarations: it is a fig leaf for bad design, and it does not prevent anyone from using the declaration---and experience has shown that they will. If the method breaks the encapsulation of its class, then its internal representation will be hard to change. Instead, use [RestrictedApiChecker](http://errorprone.info/bugpattern/RestrictedApiChecker), which enforces fine-grained visibility policies.

SimonCockx (approved):

PR #730, opened 2024-02-25
Validation result

hugohills-regnosys:

## Type of change


SimonCockx (approved):

SimonCockx (approved):

rosetta-runtime/src/main/java/com/rosetta/model/lib/meta/RosettaMetaData.java
```diff
@@ -16,12 +15,6 @@ public interface RosettaMetaData<T extends RosettaModelObject> {
 
 	List<Validator<? super T>> dataRules(ValidatorFactory factory);
 	
-	// @Compat. This will be empty, as choice rules are now all data rules.
```
hugohills-regnosys, line 19: Won't removing this cause backward compat issues with generated code?

SimonCockx: This has been out of use (not being generated) from even before blueprints, so no old generated code should contain it anymore.

PR #732, opened 2024-03-07
task: 4188 Disabled the expectations cache

minesh-s-patel:
Disabled the expectations cache as it is causing an issue in Rosetta where some ecore eproxies are being incorrectly cached.


rosetta-lang/src/main/java/com/regnosys/rosetta/types/RosettaTypeProvider.xtend
```diff
@@ -186,6 +191,9 @@ class RosettaTypeProvider extends RosettaExpressionSwitch<RType, Map<EObject, RT
 		])
 	}
 	private def RType getRTypeFromCache(String cacheKey, EObject object, Provider<RType> typeProvider) {
+		if (!EXPRESSION_RTYPE_CACHE_KEY_ENABLED) {
+			typeProvider.get()
```
SimonCockx, line 195: Shouldn't there be a `return` here?

minesh-s-patel: Actually - you are right. I take it back. Adding the return!

PR #761, opened 2024-05-23
Added overridden syntax to allow rosetta files to flag when they fork…

minesh-s-patel:
Added overridden syntax to allow rosetta files to flag when they fork a parent model file

## Type of change

### DIFF
--- a/rosetta-ide/rosetta.tmLanguage.yaml
+++ b/rosetta-ide/rosetta.tmLanguage.yaml
@@ -57,7 +57,7 @@ repository:
     patterns:
     - include: '#namespace'
     - include: '#version'
-    - include: '#overridden'
+    - include: '#override'
     - include: '#import'
     - include: '#configuration'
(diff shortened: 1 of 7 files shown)


rosetta-ide/rosetta.tmLanguage.yaml
```diff
@@ -88,6 +89,16 @@ repository:
...
+      override:
+        name: meta.override.rosetta
+        begin: '{{wordStart}}override{{wordEnd}}'
+        beginCaptures:
+          0: { name: keyword.other.override.rosetta }
+        end: '{{rootEnd}}'
+        patterns:
+        - include: '#comment'
+        - include: '#string'
+
```
SimonCockx, line 101: Could be better integrated inside `import`, but how about an already slight improvement?
```
      override:
        name: keyword.other.override.rosetta
        match: '{{wordStart}}override{{wordEnd}}'
```

PR #850, opened 2024-10-01
Implemented choice switch

SimonCockx:
## Type of change

### DIFF
--- a/docs/rune-modelling-component.md
+++ b/docs/rune-modelling-component.md
@@ -912,7 +912,7 @@ Rune supports basic arithmetic operators
 
 The `default` operator takes two values of matching type. If the value to the left of the default is empty then the value of to the right of the default will be returned. Note that the type and cardinality of both sides of the operator must match for the syntax to be valid.
 
-### Switch Operator
+#### Switch Operator
 
 The `switch` operator takes as its left hand input an argument on which to perform case analysis. The right side of the operator takes a set of case statements which define a return value for the expression when matching that case to the input.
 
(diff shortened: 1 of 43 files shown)


docs/rune-modelling-component.md
SimonCockx, line 1: Work in progress

hugohills-regnosys (approved):

docs/rune-modelling-component.md
```diff
@@ -912,31 +912,69 @@ Rune supports basic arithmetic operators
...
 ```
 
 {{< notice info "Note" >}}
-The `default` case is optional, in case when there is no match then the `empty` value is returned.
+The `default` case is optional. In case when there is no match then the `empty` value is returned.
 {{< /notice >}}
 
+##### `switch` for `choice` types
+
+Consider the following model for a `Vehicle`.
```
hugohills-regnosys, line 942: Please add these to the Demo project

rosetta-lang/src/main/java/com/regnosys/rosetta/validation/RosettaSimpleValidator.xtend
```diff
@@ -172,57 +176,99 @@ class RosettaSimpleValidator extends AbstractDeclarativeRosettaValidator {
 			warning(msg, owner, ref, index)
 		}
 	}
-
+	
 	@Check
-	def void switchInputsMustBeSingleCardinality(SwitchOperation op) {
+	def void checkSwitch(SwitchOperation op) {
```
davidalk, line 181: Maybe we could split these into three private methods

davidalk (approved):

docs/rune-modelling-component.md
Oblongs, line 942: I find this example overly complex and therefore difficult to understand.

Could you simplify the example to show a non-nested and syntactically correct use.  For example:

```
choice Vehicle:
  PetrolCar
  ElectricCar

type Car:
  fuelCapacity number (1..1)

type ElectricCar:
  batteryCapacity number (1..1)
```

and then a simple use, eg:

```
set efficiency:  vehicle switch 
  ElectricCar        then 5 * batteryCapacity // assume 5 kilometres per kWh of battery
  Car                then 15 * fuelCapacity   // assume 15 kilometres per litre of fuel
```

Issue #858, opened 2024-10-23
Automation of code formatting for consistency


maria77102:
The Rosetta IDE contains a ‘format code’ function, but since ‘beautifying’ the code is not mandatory, different contributions to one document can end up looking inconsistent- making it hard to read/review.

The goal is to create a standard formatting, which is going to be applied automatically after every contribution is submitted. This way, complex code is easier to read/review.

High level strategy:

- Create a new service which accepts a collection of documents and returns a collection of formatted resources
- Implement the interface
- Create a Maven plugin to automate the formatting process - applying it after every contribution
- Test the implementation to identify any issues and address them to ensure consistent code formatting.

PR #859, opened 2024-10-28
Automating Formatting of Rune Code

maria77102:
This PR deals with issue #858.

## Type of change


rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/CodeFormatterService.java
```diff
@@ -0,0 +1,10 @@
+package com.regnosys.rosetta.formatting2;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.eclipse.emf.ecore.resource.Resource;
+
+public interface CodeFormatterService {
+	List<Resource> formatCollection(List<Resource> resources) throws IOException;
```
SimonCockx, line 9: If this method modifies the resources in-place, I would change this to `void`. A bit of documentation to explain this behaviour would also be good.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/CodeFormatterService.java
```diff
@@ -0,0 +1,10 @@
+package com.regnosys.rosetta.formatting2;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.eclipse.emf.ecore.resource.Resource;
+
+public interface CodeFormatterService {
```
SimonCockx, line 8: This name is perhaps a bit generic - it doesn't really tell you what it does. What about `ResourceFormatterService`?

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RuneFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+	@Override
+	public List<Resource> formatCollection(List<Resource> resources) throws IOException {
+		
+		List<Resource> result = new ArrayList<Resource>();
+		
+		for (Resource doc: resources) {
+			Resource formattedDocument = formatDocument(doc);
+			result.add(formattedDocument);
+		}	
+		return result;
```
SimonCockx, line 45: This is perfectly fine, but just as a challenge, could you rewrite this using Java stream API? :)

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RuneFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+public class RuneFormatter implements CodeFormatterService{
+	@Inject
+	private Provider<FormatterRequest> formatterRequestProvider;
+	
+	@Inject 
+	private Provider<RosettaFormatter> rosettaFormatterProvider;
```
SimonCockx, line 31: This one is injecting the implementation rather than the interface. Change to inject the formatter interface instead.

Just to explain why: this is against the principle of dependency injection, since it breaks the separation between interface and implementation, making the codebase less extensible.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RuneFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+		List<Resource> result = new ArrayList<Resource>();
+		
+		for (Resource doc: resources) {
+			Resource formattedDocument = formatDocument(doc);
+			result.add(formattedDocument);
+		}	
+		return result;
+	}
+
+	private Resource formatDocument(Resource doc) throws IOException {
```
SimonCockx, line 48: Since the word `Document` is used for something specific in Xtext's formatting code (e.g., `IFormattableDocument`), the name `formatDocument` might be misleading. What about `formatResource`?

SimonCockx: I wonder if we should expose this method in the interface as well. It might be good if at some point a use case arises for formatting a single resource.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RuneFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.xtext.formatting2.regionaccess.internal.TextRegionRewriter;
+import org.eclipse.xtext.nodemodel.INode;
+import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
+import org.eclipse.xtext.resource.XtextResource;
+import org.eclipse.xtext.util.TextRegion;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+public class RuneFormatter implements CodeFormatterService{
```
SimonCockx, line 26: Make name of class consistent with interface.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RuneFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+		RosettaFormatter rosettaFormatter = rosettaFormatterProvider.get();
+		
+		// Format root object in document
+		EObject content = doc.getContents().get(0);
+		
+		//figure out which region should be formatted (all of it basically)
+		INode node = NodeModelUtils.getNode(content);
+		
+		TextRegion textRegion = new TextRegion(node.getOffset(), node.getLength());
+		req.setRegions(Collections.singletonList(textRegion));
```
SimonCockx, line 60: Have a look at the documentation of `regions` inside `FormatterRequest`. I think this can be simplified.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RuneFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+		
+		ITextRegionAccess regionAccess = regionBuilder.forNodeModel((XtextResource) doc).create();
```
SimonCockx, line 62: This is interesting. I don't think we should cast here.

Just some background: in our case, all resources are `XtextResource`s for now. However, there is support for adding other kinds of resources as well to a resource set. Suppose for example that we somehow would add support for loading `.csv` files or `.xml` files or other "raw" resources, then this will crash.

Instead what I think we should do is change the input type to `XtextResource` (because other resources are not supported), and then in the method `formatCollection` filter out any non-`XtextResource`s.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RuneFormatter.java
```diff
@@ -0,0 +1,82 @@
...
+		req.setTextRegionAccess(regionAccess);
+		
+		//initialize the formatter with the request - telling it what to format
+		rosettaFormatter.initialize(req);
+		//list contains all the replacements which should be applied to doc
+		List<ITextReplacement> replacements = rosettaFormatter.format(req);
+				
+		
+		//formatting using TextRegionRewriter
+		TextRegionRewriter regionRewriter = new TextRegionRewriter(regionAccess);
```
SimonCockx, line 72: See docs on class for the recommended way of getting an instance. :)

rosetta-testing/src/test/java/com/regnosys/rosetta/formatting2/FormatterServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+@ExtendWith(InjectionExtension.class)
+@InjectWith(RosettaInjectorProvider.class)
+public class FormatterServiceTest {
+	@Inject CodeFormatterService formatterService;
+	@Inject Provider<ResourceSet> resourceSetProvider;
+	@Inject ISerializer serializer;
+	
+	@Test
+	void formatSingleDocument() throws IOException, URISyntaxException {
```
SimonCockx, line 35: Since `formatSingleDocument` and `formatMultipleDocuments` share a lot of code, I think it would improve maintainability if we added a little test utility method to this class. E.g., you pass it a list of files, and the test will format those files and compare them to their expectation.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -0,0 +1,80 @@
...
+	@Inject 
+	private Provider<IFormatter2> iFormatter2Provider;
+
+	@Inject
+	private TextRegionAccessBuilder regionBuilder;
+	
+	@Override
+	public void formatCollection(Collection<Resource> resources) {
+	    formatCollection(resources, null);
+	}
```
SimonCockx, line 36: Trivial implementations such as these are usually implemented as a [`default`](https://www.geeksforgeeks.org/default-methods-java/) interface method, so implementers of the interface do not have to write boilerplate code. I would move this one up, together with the two below.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -0,0 +1,80 @@
...
+	public void formatCollection(Collection<Resource> resources, ITypedPreferenceValues preferenceValues){
+		resources.stream().forEach(resource -> {
+			if (resource instanceof XtextResource) {
+	            try {
+	                formatXtextResource((XtextResource) resource, preferenceValues);
+	            } catch (IOException e) {
+	                e.printStackTrace();
+	            }
+	        } else {
+	            System.out.println("Resource is not of type XtextResource and will be skipped: " + resource);
```
SimonCockx, line 53: Let me introduce you to the wonderful world of logging APIs :)

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -0,0 +1,80 @@
...
+	        }
+		});
+	}
+
+	@Override
+	public void formatXtextResource(XtextResource resource, ITypedPreferenceValues preferenceValues) throws IOException {
```
SimonCockx, line 59: I'm surprised to see a `throws IOException` in this declaration, since conceptually this method does not perform any I/O. More below.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -0,0 +1,80 @@
...
+		List<ITextReplacement> replacements = formatter.format(req);
+				
+		//formatting using TextRegionRewriter
+		ITextRegionRewriter regionRewriter = regionAccess.getRewriter();
+		String formattedString = regionRewriter.renderToString(regionAccess.regionForDocument(), replacements);
+
+		//With the formatted text, update the resource
+		InputStream resultStream = new ByteArrayInputStream(formattedString.getBytes(StandardCharsets.UTF_8));
+		resource.unload();
+		resource.load(resultStream, null);
```
SimonCockx, line 78: I'm guessing you added the `IOException` to the declaration because this method potentially throws an `IOException`. This is because in general the input stream could come from anywhere, e.g., from reading a file, or from a HTTP request over the internet. However, we know that the input stream is actually coming from an in-memory string (so no I/O happening), so in practice I would not expect an `IOException` here, so I don't think we should leak it to the callers of this method.

That's a long way of saying I think we should catch the error here, wrap it in an `UncheckedIOException`, and then rethrow it. :) (together with a small comment explaining why we expect an `IOException` never to happen)

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -0,0 +1,77 @@
...
+		//formatting using TextRegionRewriter
+		ITextRegionRewriter regionRewriter = regionAccess.getRewriter();
+		String formattedString = regionRewriter.renderToString(regionAccess.regionForDocument(), replacements);
+
+		//With the formatted text, update the resource
+		InputStream resultStream = new ByteArrayInputStream(formattedString.getBytes(StandardCharsets.UTF_8));
+		resource.unload();
+		try {
+			resource.load(resultStream, null);
+			resource.save(null);
```
SimonCockx, line 71: This is interesting. I think it would be good to keep the formatting in-memory only, and then users of the class can decide whether they wanna actually save it or not. That way it might cover more use cases.

So what I would do is move this `save` into your main class, and perhaps also document that the formatting is done in-memory.

rosetta-tools/src/main/java/com/regnosys/rosetta/tools/ResourceFormattingTool.java
```diff
@@ -0,0 +1,80 @@
...
+ * <h2>Usage:</h2>
+ * <pre>
+ * java ResourceFormattingTool /path/to/directory
+ * </pre>
+ * <p>
+ * If no valid directory path is provided as an argument, the program will exit with an error message.
+ * </p>
+ */
+public class ResourceFormattingTool {		
+	private static Logger LOGGER = LoggerFactory.getLogger(XtextResourceFormatter.class);
```
SimonCockx, line 39: This one should be `LoggerFactory.getLogger(ResourceFormattingTool.class)`

SimonCockx (approved):

PR #864, opened 2024-11-07
Formatter Maven plugin

maria77102:
This PR deals with the 4th point of issue #858.

## Type of change


rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaExpressionFormatter.xtend
```diff
@@ -250,11 +250,13 @@ class RosettaExpressionFormatter extends AbstractRosettaFormatter2 {
 	}
 	
 	private def dispatch void unsafeFormatExpression(RosettaSymbolReference expr, extension IFormattableDocument document, FormattingMode mode) {
+		val extension referenceCallGrammarAccess = rosettaReferenceOrFunctionCallAccess
```
SimonCockx, line 253: ooh, nice one finding the right rule name for that one!

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.xtend
```diff
@@ -236,6 +237,18 @@ class RosettaFormatter extends AbstractRosettaFormatter2 {
 		ele.regionFor.assignment(nameAssignment_1)
 			.prepend[oneSpace]
 	}
+	
+	def dispatch void format(Choice ele, extension IFormattableDocument document) {
+		ele.regionFor.keyword('choice')
```
SimonCockx, line 242: I think it is good practice to use string-based keywords as little as possible, since they are a bit error prone. E.g., if we change the keyword to be `"union"` instead of `"choice"` in the parser, we wouldn't notice that this needs changing. However, by using `choiceAccess.choiceKeyword_0` (or whatever it is), you would be notified by an error in the IDE, since `choiceKeyword_0` doesn't exist anymore.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.xtend
```diff
@@ -236,6 +237,18 @@ class RosettaFormatter extends AbstractRosettaFormatter2 {
...
+			.append[oneSpace]
+		ele.regionFor.keyword(':')
+			.prepend[noSpace]
```
SimonCockx, line 245: Similar to `Data`, a `Choice` can also have a definition (documentation), so it would be good to add `ele.formatDefinition(document)` too.

SimonCockx: That'll make sure that `<"my docs">` are formatted correctly. I.e., not as `  <    "my docs"  >  `

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.xtend
```diff
@@ -236,6 +237,18 @@ class RosettaFormatter extends AbstractRosettaFormatter2 {
...
+		ele.indentInner(document)
+		ele.attributes.forEach[
+			prepend[newLine]
+			format
+		]
```
SimonCockx, line 250: To have a consistent formatting for newlines, I think we should have the same kind of logic here as for `Data`.

```
		ele.attributes.head
			.prepend[setNewLines(1, 2, 2)]
			.format
		ele.attributes.tail.forEach[
			prepend[newLine]
			format
		]
```
The first case makes sure the first attribute is either prepended with one newline or with two newlines. By default it is two.
The second case makes sure all subsequent attributes are prepended by one newline.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -48,16 +50,37 @@ public void formatCollection(Collection<Resource> resources, ITypedPreferenceVal
...
+
 		req.setTextRegionAccess(regionAccess);
 
 		// list contains all the replacements which should be applied to resource
-		List<ITextReplacement> replacements = formatter.format(req);
+		List<ITextReplacement> replacements = new ArrayList<>();
+		try {
+			replacements = formatter.format(req); // throws exception
+		} catch (RuntimeException e) {
+			LOGGER.error("RuntimeException in " + resource.getURI() + ": " + e.getMessage());
```
SimonCockx, line 75: When logging an error, always pass the error to the logger as well. I.e.,
```
LOGGER.error(msg, e);
```
=> the logger can then be configured whether or not to show the stack trace.

rosetta-maven-plugin/src/main/java/com/regnosys/rosetta/maven/ResourceFormatterMojo.java
```diff
@@ -0,0 +1,72 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.Injector;
+import com.regnosys.rosetta.RosettaStandaloneSetup;
+import com.regnosys.rosetta.builtin.RosettaBuiltinsService;
+import com.regnosys.rosetta.formatting2.ResourceFormatterService;
+
+@Mojo(name = "format")
+public class ResourceFormatterMojo extends AbstractMojo {
```
SimonCockx, line 27: I think this one could use a bit of documentation on its usage.

rosetta-maven-plugin/src/main/java/com/regnosys/rosetta/maven/ResourceFormatterMojo.java
```diff
@@ -0,0 +1,72 @@
...
+					resource.save(null);
+					LOGGER.info("Successfully formatted and saved file at location " + resource.getURI());
+				} catch (IOException e) {
+					LOGGER.error("Error saving file at location " + resource.getURI() + ": " + e.getMessage());
+				} catch (RuntimeException e) {
+					LOGGER.error("RuntimeException while saving in following file: " + resource.getURI() + ": "
+							+ e.getMessage());
+				}
+			});
+		} catch (IOException e) {
```
SimonCockx, line 67: I'm guessing this `catch` is for the directory walking? As a good practice I would try to limit the size of `try` blocks, so there is less danger of accidentally catching errors you didn't intend to catch. So it would be good to put this just around the `List<Resource> resources = ...`.

In fact, I wonder if we should actually be throwing an error if we fail to read a file. E.g., throw a `MojoFailureException` in the `catch`. Also interesting: https://stackoverflow.com/questions/1033154/when-to-use-mojoexecutionexception-vs-mojofailureexception-in-maven

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -48,16 +50,37 @@ public void formatCollection(Collection<Resource> resources, ITypedPreferenceVal
...
-		List<ITextReplacement> replacements = formatter.format(req);
+		List<ITextReplacement> replacements = new ArrayList<>();
+		try {
+			replacements = formatter.format(req); // throws exception
+		} catch (RuntimeException e) {
+			LOGGER.error("RuntimeException in " + resource.getURI() + ": " + e.getMessage(), e);
+		}
+
+		// Abort if replacements is empty (either because nothing to format or method
+		// throws exception
```
SimonCockx, line 79: Missing ending `)` in your comment :)

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/ResourceFormatterService.java
```diff
@@ -1,70 +1,139 @@
...
+	 * @param resources a collection of {@link XtextResource} objects to be
+	 *                  formatted
+	 * @param handler   a {@link java.util.function.BiConsumer} that processes each
+	 *                  formatted resource. The first argument is the
+	 *                  {@link Resource}, and the second is the formatted content as
+	 *                  a {@link String}.
 	 */
-	default void formatCollection(Collection<Resource> resources) {
-		formatCollection(resources, null);
+	default void formatCollection(Collection<Resource> resources, BiConsumer<Resource, String> handler) {
```
SimonCockx, line 37: Maybe it's mostly a style preference, but for readability I would prefer having a custom functional interface. E.g., `IFormattedResourceAcceptor`, which exposes a single `accept` method. That way, you could also document the inputs of the `accept` method separately. And it gives you the opportunity to define your own functional interface. ;)

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -47,30 +45,39 @@ public void formatCollection(Collection<Resource> resources, ITypedPreferenceVal
...
 		FormatterRequest req = formatterRequestProvider.get();
 		req.setPreferences(preferenceValues);
 		IFormatter2 formatter = iFormatter2Provider.get();
 
-		ITextRegionAccess regionAccess = regionBuilder.forNodeModel(resource).create();
+		ITextRegionAccess regionAccess = null;
+		try {
+			regionAccess = regionBuilder.forNodeModel(resource).create();
+		} catch (Exception e) {
+			LOGGER.info("Resource " + resource.getURI() + " is empty.", e);
```
SimonCockx, line 61: About this error message: are you sure that this will only occur if the resource is empty? Since you're capturing a general `Exception` here.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/XtextResourceFormatter.java
```diff
@@ -47,30 +45,39 @@ public void formatCollection(Collection<Resource> resources, ITypedPreferenceVal
...
+			return;
+		}
+
 		req.setTextRegionAccess(regionAccess);
 
 		// list contains all the replacements which should be applied to resource
-		List<ITextReplacement> replacements = formatter.format(req);
+		List<ITextReplacement> replacements = new ArrayList<>();
```
SimonCockx, line 68: Small optimization: put the initializer in the `catch` branch instead. That way you don't create an additional new array list if the formatting succeeds.

rosetta-maven-plugin/pom.xml
```diff
@@ -59,6 +59,16 @@
...
+		<dependency>
+			<groupId>com.regnosys.rosetta</groupId>
+			<artifactId>com.regnosys.rosetta.ide</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>com.regnosys.rosetta</groupId>
+			<artifactId>com.regnosys.rosetta.lib</artifactId>
+			<version>${project.version}</version>
+		</dependency>
```
SimonCockx, line 71: This one should actually already be included transitively by `<artifactId>com.regnosys.rosetta</artifactId>`. Are you sure you need it?

rosetta-maven-plugin/pom.xml
```diff
@@ -59,6 +59,16 @@
 			<groupId>org.eclipse.xtext</groupId>
 			<artifactId>xtext-maven-plugin</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.regnosys.rosetta</groupId>
+			<artifactId>com.regnosys.rosetta.ide</artifactId>
+			<version>${project.version}</version>
+		</dependency>
```
SimonCockx, line 66: I'm a bit surprised to see this one here. Are you sure you need it?

rosetta-maven-plugin/src/main/java/com/regnosys/rosetta/maven/ResourceFormatterMojo.java
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+		// Create an ObjectMapper, read JSON into a Map
+		ObjectMapper objectMapper = new ObjectMapper();
+		Map<String, Object> map = null;
+		try {
+			map = objectMapper.readValue(resourceStream, Map.class);
+		} catch (IOException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
```
SimonCockx, line 148: Might wanna throw a failure exception here.

rosetta-maven-plugin/src/main/java/com/regnosys/rosetta/maven/ResourceFormatterMojo.java
```diff
@@ -0,0 +1,207 @@
...
+					try {
+						Files.writeString(resourcePath, formattedText);
+						LOGGER.info("Content written to file: " + resourcePath);
+					} catch (IOException e) {
+						LOGGER.error("Error writing to file.", e);
+					}
+				});
+	}
+
+	private FormattingOptions readFormattingOptions(String options) throws IOException {
```
SimonCockx, line 131: Perhaps rename "options" to "optionsPath"?

rosetta-maven-plugin/src/main/java/com/regnosys/rosetta/maven/ResourceFormatterMojo.java
```diff
@@ -0,0 +1,201 @@
...
+			} else if (value instanceof Boolean) {
+				formattingOptions.putBoolean(key, (Boolean) value);
+			} else {
+				throw new IllegalArgumentException("Unsupported value type for key: " + key);
+			}
+		}
+		return formattingOptions;
+	}
+
+	public static ITypedPreferenceValues createPreferences(FormattingOptions options) {
```
SimonCockx, line 166: Since this code is exactly the same as in the IDE project, I think it would be good to refactor it so we use some common code. How about adding a `FormattingOptionsAdaptor` in `rosetta-lang` which does the conversion, and then that one can be injected both in IDE and in this plugin?

SimonCockx: You can then probably move the lsp4j dependency from this project to rosetta-lang as well.

rosetta-tools/pom.xml
```diff
@@ -72,6 +72,11 @@
 			<version>${project.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.regnosys.rosetta</groupId>
+			<artifactId>rosetta-maven-plugin</artifactId>
+			<version>0.0.0.main-SNAPSHOT</version>
+		</dependency>
```
SimonCockx, line 79: Given my other comment, you can also remove this dependency and inject the common adaptor instead.

rosetta-tools/src/main/java/com/regnosys/rosetta/tools/ResourceFormattingTool.java
```diff
@@ -39,42 +41,65 @@ public class ResourceFormattingTool {
...
+        
+        // check if optional parameter was given. If not use default value
+        FormattingOptions formattingOptions = null;
+        if(args.length > 1) {
+        	String formattingOptionsPath = args[1];
+        	try {
+    			formattingOptions = ResourceFormatterMojo.readFormattingOptions(formattingOptionsPath);
+    		} catch (IOException e) {
+    			LOGGER.error("Config file not found.", e);
+    		} catch (MojoFailureException e) {
```
SimonCockx, line 67: Right now this is leaking Mojo specific exceptions. See previous comment to solve it.

rosetta-ide/src/main/java/com/regnosys/rosetta/ide/server/RosettaFormattingOptionsRetriever.java
```diff
@@ -0,0 +1,13 @@
+package com.regnosys.rosetta.ide.server;
+
+import java.util.concurrent.CompletableFuture;
+
+import org.eclipse.lsp4j.FormattingOptions;
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
+import org.eclipse.lsp4j.services.LanguageServer;
+
+public interface RosettaFormattingOptionsRetriever extends LanguageServer{
```
SimonCockx, line 9: Usually something like this is called a `Provider`. :) Although, since it also extends `LanguageServer`, and we might use this interface to drop any other extension we might wanna make, I think it would be fine to just call this `RosettaLanguageServer`, as a catch-all for all future extensions.

rosetta-ide/src/main/java/com/regnosys/rosetta/ide/server/RosettaLanguageServerImpl.java
```diff
@@ -170,4 +175,15 @@ protected SemanticTokens semanticTokensRange(SemanticTokensRangeParams params, C
...
 	}
+
+	@Override
+	public CompletableFuture<FormattingOptions> getDefaultFormattingOptions() {
+		try {
+			return CompletableFuture.completedFuture(formattingOptionsAdapter.readFormattingOptions(null));
+		} catch (IOException e) {
+			// should never happen, since null path always leads to default options being returned
+			e.printStackTrace();
+			return null;
```
SimonCockx, line 186: Return a `CompletableFuture.failedFuture(e)` in this case, so the client knows if anything wrong happened.

rosetta-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaExpressionFormatter.xtend
```diff
@@ -175,15 +255,23 @@ class RosettaExpressionFormatter extends AbstractRosettaFormatter2 {
 	private def dispatch void unsafeFormatExpression(RosettaConditionalExpression expr, extension IFormattableDocument document, FormattingMode mode) {
 		val extension conditionalGrammarAccess = rosettaCalcConditionalExpressionAccess
 		
+		// fix edge case where 'then' inside constructor value is not indented correctly
+		if (expr.eContainer instanceof ConstructorKeyValuePair) {
```
SimonCockx, line 259: I guess this "check" is not necessary anymore, and you should just call `surroundIndent...`?

rosetta-ide/src/main/java/com/regnosys/rosetta/ide/server/RosettaLanguageServerImpl.java
```diff
@@ -170,4 +175,15 @@ protected SemanticTokens semanticTokensRange(SemanticTokensRangeParams params, C
...
 		return this.getRequestManager().runRead((cancelIndicator) -> this.semanticTokensRange(params, cancelIndicator));
 	}
+
+	@Override
+	public CompletableFuture<FormattingOptions> getDefaultFormattingOptions() {
+		try {
+			return CompletableFuture.completedFuture(formattingOptionsAdapter.readFormattingOptions(null));
+		} catch (IOException e) {
+			// should never happen, since null path always leads to default options being returned
+			e.printStackTrace();
```
SimonCockx, line 185: This one can go now :)

SimonCockx (approved):

PR #903, opened 2025-01-22
enable java 17 for testing project

### LINKED ISSUE #785: Caching sometimes breaks change propagation
Repro:

```
namespace foo

import bar.*

body Authority Body

corpus Regulations Corpus 

report Body Corpus in T+2
    from Event
    when Eligible
    with type Report

eligibility rule Eligible from Event:
    item

reporting rule Thing from Event:
    ""
```

```
namespace foo

type Report:
    thing string (1..1)
        [ruleReference Thing]
```

```
namespace bar

type Event:
```

1. Observe that it validates fine.
2. Change the name of `Event` to `Even`, and observe that it breaks (as expected).
3. Change the name of `Even` back to `Event`, and observe that one validation error persists on the report.

### LINKED ISSUE #786: Highlighting broken for `only exists` and `required choice` with a comma
`only exists` with multiple inputs and `required choice` (and `optional choice`) break highlighting for everything below.

Examples from the CDM: `ProductTaxonomy`

### LINKED ISSUE #844: `MetaFieldGenerator` is generating too much Java files
In a recent PR, a condition has accidentally slightly changed, which is causing superfluous Java files to be generated. This causes longer compilation times.

See https://github.com/finos/rune-dsl/commit/c20eb67dc6388d7ec3a33ddce6c084d41933b381#diff-8009a8dbaf539bf5ba5c94ea3b47e647586227addce959d73e56fd8637e33f0eL110
condition `hasMetas` (also on line 92).

minesh-s-patel:
- Moved integration tests using J17


rosetta-integration-tests/src/main/java/com/regnosys/rosetta/Main.java
SimonCockx, line 1: This seems unnecessary?

rosetta-lang/src/main/java/com/regnosys/rosetta/GenerateRosetta.mwe2
```diff
@@ -31,10 +31,10 @@ Workflow {
 				}
 				runtimeTest = {
```
SimonCockx, line 32: I think this should actually refer to the test utility project. It's where Xtext will generate it's test utilities.

minesh-s-patel: When the MWE2 runtimeTest points to `rosetta-testing`
The issue is that it generates 2 things

1. A Unit test (in `rosetta-testing/src/test`) - which is a Hello World test intended to be overriden.
2. A type called RosettaInjectorProvider.java which (in `rosetta-testing/src/main`) which we need in the testing utils

When I just moved the all the tests to integration testing - it started to generate point 1 which caused a unit test failure.

When I move the MWE2 runtimeTest to point to `rosetta-integration-tests`, then we are generating point 2 int he wrong place.

I made a change to just disable the runtime test automation. The automation is meant to be a starting point and does not really (I think) provide much value any more.

The other option is that we move the MWE2 runtimeTest back to `rosetta-testing`, and check in the generated xtend test and live with 1 test being in rosetta-testing. I prefer the first option.  @SimonCockx - take a look at the PR now and tell me what you think.

SimonCockx (approved):

Issue #879, opened 2024-12-03
Loading External Reference Data into CDM - FpML Coding Schemes


mgratacos:
### Problem Statement
CDM currently loads FpML coding schemes and compiles them generating a static enumerated list using the Rune DSL `docReference` annotations. 

- This creates extra work for maintaining and releasing the CDM model.
- This causes issues with current CDM users since these FpML coding schemes are updated more frequently than the CDM users upgrade their CDM implemented code.
- FpML based code lists have changed an average of monthly for over a decade (140+ changes over 11 years).
- This implies 140+ releases of CDM for reference data updates.
- If multiple major versions of CDM are supported at once, this may imply several times more releases of CDM.
- Rapidly evolving reference data exacerbates the challenge (e.g. large commodity code lists).
- Inefficiencies limit overall effectiveness, inhibit interoperability and have more impact as CDM grows in adoption. 

### Lists in Scope
**There are currently 15 CDM enumerations automatically regenerated by CDM** on each release based on the FpML coding scheme.  These all should be converted to the new mechanism.

**10 CDM enumerated lists look like good candidates to be moved to the new mechanism**  because the schemes are relatively large (10+ values) and are expected to change relatively frequently, and the values are used infrequently in the code in most cases. However, they require further analysis and discussion.

### Proposed Implementation
The proposed implementation changes the Rune syntax by extending the `string` data type (details subject to refinement based on detailed design and implementation work and feedback from the CDM TAWG and Rune working groups):
- The existing Rune `string` type would be enhanced to include a new user-defined validation mechanism.
  - The existing `string` type supports validated properties (constraints) for `minLength`, `maxLength`, and `pattern`.
  - Two new constraint properties would be added:
    - `validationRule` would be a string that identifies the name of a user-defined validation rule that would be executed when the `string` is set, similar to the existing validation constraints.  For the purposes of this project, we would always set this to the same value, something like "`CodeListValidation`”. Different values of this property could be used for other types of validation, such as CRCs/CheckSums, or database or API lookups.
    - `domain` would be a string that identifies the list of values to validate against, e.g. “currency-code” or “business-center” or “floating-rate-index”, and could be used by the validation rule to refine its validation.
  - The validation generation logic that currently validates the length and pattern constraints would be enhanced to invoke a validation stub function `ValidateString` with the parameters `validationRule`, `domain`, and the value of the string to be validated.
  - In the Rune runtime, this validation stub function would simply return a warning to implement the validation function.
[...]

PR #909, opened 2025-01-27
Adding condition support in typeAlias

manel-martos:
This release addresses the requirements specified in issue [#879](https://github.com/finos/rune-dsl/issues/879) by introducing updates to the `RosettaTypeAlias` functionality. Key changes include enhancements to the abstract syntax tree, parser rules, and implicit variable handling, along with corresponding test coverage.

## New feature

- Added `conditions` property to the `RosettaTypeAlias` syntax tree.
- Introduced a new `Condition` parser rule in `RosettaTypeAlias`.
- Added implicit variable `item` at the typeAlias condition level in `ImplicitVariableUtil`.
- Included test coverage for the new syntax in `RosettaParsingTest.xtend`.


manel-martos:
@minesh-s-patel @SimonCockx
@minesh-s-patel @SimonCockx
The contribution has been upgraded to modify type-format validator generation by introducing a `private final List<Condition> conditions` field in `RAliasType`. The new generation code is triggered only for `typeAliases` with a condition named `IsValidCodingScheme`. The generated code now references the function defined in the extended CDM model, and we likely need to instantiate the default NoOp implementation.  

The function definition in the extended CDM model is as follows:  

```
func ValidateFpMLCodingSchemeDomain:
	inputs:
		code FpMLCodingScheme (0..1)
		domain string (1..1)
	output:
		isValid boolean (1..1)
```

We have tested the generated `rune-dsl` against our extended CDM model with the following object structure:  

```			
typeAlias FpMLCodingScheme(domain string):
	string
	
	condition IsValidCodingScheme: 
		ValidateFpMLCodingSchemeDomain (item, domain)

typeAlias BusinessCenter: FpMLCodingScheme(domain: "business-center")
typeAlias FloatingRateIndex : FpMLCodingScheme(domain: "floating-rate-index")
typeAlias Currency : FpMLCodingScheme(domain: "currency-code")
```

An example of the generated code that runs as expected:  
```
//GEM-TH: cdm-ref-data validation mock impl
	protected cdm.base.staticdata.codelist.functions.ValidateFpMLCodingSchemeDomain func = new cdm.base.staticdata.codelist.ValidateFpMLCodingSchemeImpl();

	private List<ComparisonResult> getComparisonResults(BusinessCenterTime o) {
		return Lists.<ComparisonResult>newArrayList(
				func.evaluate(o.getBusinessCenter().getValue(), "business-center")?
					ComparisonResult.success() : ComparisonResult.failure(o.getBusinessCenter().getValue() + " code not found in domain 'business-center'")
			);
	}
```

The model function and object names are not yet definitive. Looking forward to your feedback to improve the contribution and ensure it does not impact other Rune models.

SimonCockx:
Hey @manel-martos, thanks for the explanation, I'll be reviewing this now! I've noticed you are currently behind the main finos branch, and have some conflicts. Could you try and resolve those please?

rosetta-lang/src/main/java/com/regnosys/rosetta/generator/java/object/ValidatorsGenerator.xtend
```diff
@@ -202,9 +220,51 @@ class ValidatorsGenerator {
...
+			null
+		} else {
+			val prop = javaType.findProperty(attr.name)
+			val propCode = prop.applyGetter(JavaExpression.from('''o''', javaType));
+			'''
+			«FOR cond : conditions»
+				«IF cond.getName().equalsIgnoreCase("IsValidCodingScheme")»
+					«IF attr.isMulti»
+						«IF !attr.RMetaAnnotatedType.hasMeta»
+							(ComparisonResult) «method(Optional, "ofNullable")»(«propCode»).orElse(«method(Collections, "emptyList")»())
```
SimonCockx, line 239: Could you please put class names such as `ComparisonResult` between `«»` characters? That'll automatically make sure that they are imported in the generated code.

manel-martos: Updated the generated class names and static methods

rosetta-lang/src/main/java/com/regnosys/rosetta/types/TypeSystem.java
```diff
@@ -299,8 +304,11 @@ public RType keepTypeAliasIfPossible(RType t1, RType t2, BiFunction<RType, RType
 			if (alias1.getTypeFunction().equals(alias2.getTypeFunction())) {
 				RTypeFunction typeFunc = alias1.getTypeFunction();
 				RType underlier = keepTypeAliasIfPossible(alias1.getRefersTo(), alias2.getRefersTo(), combineUnderlyingTypes);
+				//GEM-TH: cdm-ref-data - concatenate typeAliases existing conditions
+				List<Condition> conditions = Stream.concat(alias1.getConditions().stream(), alias2.getConditions().stream())
+	                    .collect(Collectors.toList());
 				return typeFunc.reverse(underlier)
-					.<RType>map(args -> new RAliasType(typeFunc, args, underlier))
+					.<RType>map(args -> new RAliasType(typeFunc, args, underlier, conditions))
```
SimonCockx, line 311: Same as above: should be just the conditions of the first alias.

SimonCockx (commented):
Completed the review. This is looking exceptionally clean for a first contribution, thank you for taking the time to work with the framework!

I just want to highlight that, functionally, this change currently hard-codes support for the `IsValidCodingScheme` condition. This means, while syntactically conditions on type aliases are fully supported, they won't actually do anything except for your specific use case. Given the discussions we had on this topic, I think this is reasonable, although I think for transparency we should highlight this to ISDA so they are aware which work is required to complete the feature properly. I will be raising some issues today for this.

Could you also please verify that a condition with any other name does not break the model or generated Java code? Looking at your generator, I think you might end up with syntax errors in the generated code if you define a type alias with two or more conditions. If it is not supported, at least it should also not break anything either. I suspect your logic in `checkTypeAliasFormat` might have to change a little bit so it always returns `null` in those cases instead of generating an empty string. It would be useful to add a test for this. (perhaps have a look at `ModelMetaGeneratorTest` for a similar test)

Other than that, one minor comment: I've seen a couple of comments of the form `//GEM-TH: ...`. We currently don't use this convention for attribution for the simple reason that it decreases code readability - we use native (blame) support from Git/GitHub instead. Could you please remove them, or reword them if you think they contribute to the readability of the code?

manel-martos:
> Completed the review. This is looking exceptionally clean for a first contribution, thank you for taking the time to work with the framework!
> 
> I just want to highlight that, functionally, this change currently hard-codes support for the `IsValidCodingScheme` condition. This means, while syntactically conditions on type aliases are fully supported, they won't actually do anything except for your specific use case. Given the discussions we had on this topic, I think this is reasonable, although I think for transparency we should highlight this to ISDA so they are aware which work is required to complete the feature properly. I will be raising some issues today for this.
> 
> Could you also please verify that a condition with any other name does not break the model or generated Java code? Looking at your generator, I think you might end up with syntax errors in the generated code if you define a type alias with two or more conditions. If it is not supported, at least it should also not break anything either. I suspect your logic in `checkTypeAliasFormat` might have to change a little bit so it always returns `null` in those cases instead of generating an empty string. It would be useful to add a test for this. (perhaps have a look at `ModelMetaGeneratorTest` for a similar test)
> 
> Other than that, one minor comment: I've seen a couple of comments of the form `//GEM-TH: ...`. We currently don't use this convention for attribution for the simple reason that it decreases code readability - we use native (blame) support from Git/GitHub instead. Could you please remove them, or reword them if you think they contribute to the readability of the code?

@SimonCockx - the issue has been addressed with the latest upgrade. The generator now checks condition names and expected argument presence to improve filtering and resolve the compilation errors for single and multi-condition type aliases.

Tested against the following model example:
`
type Foo:
    value1 FooAlias(fooParam: "placeholder") (1..1)
    value2 FooAlias (1..1)

typeAlias FooAlias (fooParam string): string
    condition ConditionName1: 
        fooParam exists
    condition ConditionName2: //Always fails. Current version does not generates these as a type-format validator. test for update/validatorinject branch
        False
`

SimonCockx (approved):

Issue #847, opened 2024-09-24
Provide a way to modify metadata properties


davidalk:
This proposal suggests a way we can modify a property that has metadata on it. Currently any properties that have metadata on them do not let you set that metadata using a set operator for example. The following proposed syntax would allow you to do that task:

```Haskell
func MyFunc:
	output:
		result string (1..1)
			[metadata scheme]
	set result:  "someValue"
	set result -> scheme: "someScheme"	
```

In addition to the above syntax we could also add a new `with-meta` operator which would allow you to make use of the constructor syntax when setting metadata. This would be especially useful when setting multiple metadata properties, see the following example:

```Haskell
func MyFunc:
	output:
		result string (1..1)
			[metadata scheme]
                        [metadata reference]
	set result:  "someValue"
	set result:  "a" with-meta {
				  scheme: "someScheme",
                                  reference: "someReference"             
			    }
```

PR #914, opened 2025-01-30
Story/3511 setting meta in nested objects

davidalk:
This PR implements the ability to set meta where it is nested in objects using the set/add operators as outlined in the first section of https://github.com/finos/rune-dsl/issues/847


SimonCockx (approved):

hugohills-regnosys:
https://github.com/finos/common-domain-model/issues/3364

PR #926, opened 2025-02-18
use constants for rune property names

davidalk:


rosetta-runtime/src/main/java/com/rosetta/model/lib/JavaPropertyNameConstants.java
```diff
@@ -0,0 +1,15 @@
...
+	
+	public static final String RUNE_DATA = "@data";
+	public static final String RUNE_REFERENCE = "@ref";
+	public static final String RUNE_EXTERNAL_REFERENCE = "@ref:external";
+	public static final String RUNE_SCOPED_REFERENCE = "@ref:scoped";
+	public static final String RUNE_META = "meta";
+	public static final String RUNE_SCHEME = "@scheme";
+	public static final String RUNE_KEY = "@key";
+	public static final String RUNE_EXTERNAL_KEY = "@key:external";
+	public static final String RUNE_SCOPED_KEY = "@key:scoped";
```
SimonCockx, line 14: Is the prefix on these necessary? What about putting that in the class instead?

SimonCockx (approved):

Issue #927, opened 2025-02-18
Type format validators and cardinality validators should be injected


SimonCockx:
Right now, the type format validator and the cardinality validator of a class are instantiated directly by the corresponding `Meta` class of a pojo. This results in less configurable generated code (e.g., consumers cannot override the behaviour of some validators) and makes it hard to implement future enhancements which may require dependency injection such as supporting conditions on type aliases. (see https://github.com/finos/rune-dsl/issues/879)

I would propose to inject them in the same way as we currently inject conditions, i.e., using a `ValidatorFactory` argument.

Below is an example from the CDM.

Current meta code for type `BondReference`:
``` java
@RosettaMeta(model=BondReference.class)
public class BondReferenceMeta implements RosettaMetaData<BondReference> {

	@Override
	public List<Validator<? super BondReference>> dataRules(ValidatorFactory factory) {
		return Arrays.asList(
			factory.<cdm.product.asset.BondReference>create(cdm.product.asset.validation.datarule.BondReferenceBondUnderlier.class)
		);
	}
	
	...

	@Override
	public Validator<? super BondReference> validator() {
		return new BondReferenceValidator();
	}

	@Override
	public Validator<? super BondReference> typeFormatValidator() {
		return new BondReferenceTypeFormatValidator();
	}

        ...
}
```
This would change to:
``` java
@RosettaMeta(model=BondReference.class)
public class BondReferenceMeta implements RosettaMetaData<BondReference> {

	@Override
	public List<Validator<? super BondReference>> dataRules(ValidatorFactory factory) {
		return Arrays.asList(
			factory.<BondReference>create(BondReferenceBondUnderlier.class)
		);
	}
	
	...

	@Override
	public Validator<? super BondReference> validator(ValidatorFactory factory) {
		return factory.<BondReference>create(BondReferenceValidator.class);
	}

	@Override
	public Validator<? super BondReference> typeFormatValidator(ValidatorFactory factory) {
		return factory.<BondReference>create(BondReferenceTypeFormatValidator.class);
	}

        // For backwards compatibility
	@Deprecated
	@Override
	public Validator<? super BondReference> validator() {
		return new BondReferenceValidator();
	}

	@Deprecated
	@Override
	public Validator<? super BondReference> typeFormatValidator() {
		return new BondReferenceTypeFormatValidator();
	}

        ...
}
```

PR #931, opened 2025-02-21
typeformat and cardinality validators injection

manel-martos:
This PR addresses #927 and introduces foundational improvements to support dynamic validation injection and integration.  

### New Features  
- Implements default `validator` and `typeFormatValidation` methods in **RosettaMetaData.java** for seamless integration.  
- Enhances **ModelMetaGenerator.xtend** to generate support for these methods.  
- Upgrades the coding-scheme validation function in **ValidatorsGenerator.xtend** to leverage Guice injection.  

Additional tests or documentation updates may be required. This release requires further integration work with [rune-common](https://github.com/finos/rune-common)


rosetta-runtime/src/main/java/com/rosetta/model/lib/meta/RosettaMetaData.java
```diff
@@ -33,9 +33,20 @@ public interface RosettaMetaData<T extends RosettaModelObject> {
 	
 	List<Function<? super T, QualifyResult>> getQualifyFunctions(QualifyFunctionFactory factory);
 	
+	default Validator<? super T> validator(ValidatorFactory factory) {
+		return null;
```
SimonCockx, line 37: Instead of returning null here, could you please return the deprecated `validator()`? That'll make integration a bit easier - as well as the eventual removal of the deprecated methods.

rosetta-runtime/src/main/java/com/rosetta/model/lib/meta/RosettaMetaData.java
```diff
@@ -33,9 +33,20 @@ public interface RosettaMetaData<T extends RosettaModelObject> {
...
+	}
+	
+	// @Compat. The default can be removed once validation/ingestion is in the BSP.
+	default Validator<? super T> typeFormatValidator(ValidatorFactory factory) {
+		return null;
```
SimonCockx, line 42: Same here

rosetta-runtime/src/main/java/com/rosetta/model/lib/meta/RosettaMetaData.java
```diff
@@ -33,9 +33,20 @@ public interface RosettaMetaData<T extends RosettaModelObject> {
...
+	}
+	
+	@Deprecated
 	Validator<? super T> validator();
 	
 	// @Compat. The default can be removed once validation/ingestion is in the BSP.
+	@Deprecated
```
SimonCockx, line 49: Also please add in a comment: `// since 9.37.0 - can be removed if models are all upgraded to this version or later`. I'm experimenting with better ways to track code that needs to be removed in the future - I think this can help.

manel-martos: I propose the following comment so it appears in the javadocs:
```
/**
 * @deprecated Since 9.37.0 - This method is deprecated and can be removed if models are all upgraded to this version or later.
*             Use {@link #validator(ValidatorFactory)} instead.
*/
@Deprecated
Validator<? super T> validator();

/**
* @deprecated Since 9.37.0 - This method is deprecated and can be removed if models are all upgraded to this version or later.
 *             Use {@link #typeFormatValidator(ValidatorFactory)} instead.
 */
@Deprecated
default Validator<? super T> typeFormatValidator() {
	return null;
}
```

The @Compat comments haven't been removed.

SimonCockx: Sounds good!

manel-martos: Tested `@Deprecated` metadata (`since = "9.37.0", forRemoval = false`), confirming it only works in Java 9+, while ensuring version details remain in Javadocs for broader compatibility.

PR #935, opened 2025-02-25
Handle empty inputs when set on meta function output

davidalk:


hugohills-regnosys (approved):

hugohills-regnosys (approved):

hugohills-regnosys:
https://github.com/finos/common-domain-model/issues/3364

PR #942, opened 2025-03-04
fix setting enum value on meta fields

davidalk:


rosetta-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/ExpressionGenerator.xtend
```diff
@@ -284,7 +284,15 @@ class ExpressionGenerator extends RosettaExpressionSwitch<JavaStatementBuilder,
...
-		return JavaExpression.from('''«itemType».«feature.convertValue»''', itemType)
+		return switch itemType {
+			RJavaWithMetaValue: {
+				val valueType = itemType.valueType
+				return JavaExpression.from('''«itemType».builder().setValue(«valueType».«feature.convertValue»)''', itemType)
+			}
+			default: {
+				JavaExpression.from('''«itemType».«feature.convertValue»''', itemType)
+			}
+		}
```
SimonCockx, line 295: I would suggest a bit of a different approach. What you are doing here is the job of the coercion service.

Looking at the code it used to run, the error is that the type of the returned Java expression is wrong. The only fix necessary is that `val itemType = expectedType.itemType` should actually unpack the value of meta types, and then everything is well.

SimonCockx: It might be worth adding a utility method `getValueType` in the same place as `itemType` is defined.

davidalk: Good shout, have updated

SimonCockx (approved):

PR #971, opened 2025-04-03
fix process paths for address and location

davidalk:


rosetta-runtime/src/main/java/com/rosetta/model/metafields/MetaFields.java
```diff
@@ -103,11 +101,9 @@ interface MetaFieldsBuilder extends MetaFields, RosettaModelObjectBuilder, Globa
 		default void process(RosettaPath path, BuilderProcessor processor) {
 			processor.processBasic(path.newSubPath("scheme"), String.class, getScheme(), this, AttributeMeta.META);
 			processor.processBasic(path.newSubPath("template"), String.class, getTemplate(), this, AttributeMeta.META);
-			processor.processBasic(path.newSubPath("location"), String.class, getLocation(), this, AttributeMeta.META);
-			processor.processBasic(path.newSubPath("address"), String.class, getAddress(), this, AttributeMeta.META);
+			processor.processBasic(path.newSubPath("location"), String.class, getScopedKey(), this, AttributeMeta.META);
 			processor.processBasic(path.newSubPath("globalKey"), String.class, getGlobalKey(), this, AttributeMeta.META);
 			processor.processBasic(path.newSubPath("externalKey"), String.class, getExternalKey(), this, AttributeMeta.META);
-			processRosetta(path.newSubPath("key"), processor, Key.KeyBuilder.class, getKey());
```
davidalk, line 110: This removal is the only controversial aspect as the first element of the key array is populated with the location. It's not required I can make things work without removing the `Key` processor, I just removed it for the sake of cleaning up junk. Have tested on `CDM` and `DRR` but can easily bring this key processor back if people think this is problematic.

rosetta-runtime/src/main/java/com/rosetta/model/metafields/MetaFields.java
```diff
@@ -74,11 +74,9 @@ default Class<? extends MetaFields> getType() {
 	default void process(RosettaPath path, Processor processor) {
 		processor.processBasic(path.newSubPath("scheme"), String.class, getScheme(), this, AttributeMeta.META);
```
minesh-s-patel, line 75: Do you think we should take this opportunity to make these strings into static final bars?

davidalk: done

rosetta-runtime/src/main/java/com/rosetta/model/metafields/MetaFields.java
```diff
@@ -42,7 +38,9 @@ public interface MetaFields extends RosettaModelObject, GlobalKeyFields, MetaDat
 	/*********************** Getter Methods  ***********************/
 	String getScheme();
 	String getTemplate();
+	@Deprecated
 	String getLocation();
+	@Deprecated
```
SimonCockx, line 43: I thought it would be `getLocation()` and `getKey()` that would be deprecated, in favour of `getScopedKey()`? I don't think `address` should be deprecated.

PR #1101, opened 2025-09-18
With meta set key

davidalk:


rosetta-lang/src/main/java/com/regnosys/rosetta/types/RMetaAnnotatedType.java
```diff
@@ -34,9 +34,13 @@ public RType getRType() {
...
 	}
 
-	public boolean hasMeta() {
-		return !metaAttributes.isEmpty();
-	}	
+	public boolean hasAttributeMeta() {
+		return metaAttributes.stream().anyMatch(x -> !"key".equals(x.getName()));
+	}
+	
+	public boolean hasKeyMeta() {
```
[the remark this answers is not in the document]
SimonCockx: Hm, for now I think it's alright. I know what it does when I read it

rosetta-lang/src/main/java/com/regnosys/rosetta/types/RosettaTypeProvider.java
```diff
@@ -180,9 +186,18 @@ private RMetaAnnotatedType safeRType(RosettaFeature feature, EObject context, Ma
             if (tf.getTypeCall() == null) {
                 return builtins.NOTHING_WITH_ANY_META;
             }
+
+            TypeCall typeCall = tf.getTypeCall();
+
+            List<RMetaAttribute> rMetaAttributes = new ArrayList<>();
+            if (typeCall.getType() instanceof Data typeCallType) {
+            	rMetaAttributes.addAll(getRMetaAttributesOfType(typeCallType));
+            }
```
SimonCockx, line 195: I wonder if this logic shouldn't be included inside `getRMetaAttributesOfFeature(...)`. Since that method is public, it feels inconsistent that a callee of that method will not get the key metadata in the type.

davidalk: Have moved this logic into getRMetaAttributesOfSymbol which looks to be the lowest level of call in that stack, the inner calls such as `getRMetaAttributes` and `getRMetaAttributesOfType` (the last one I created) are both util calls

rosetta-lang/src/main/java/com/regnosys/rosetta/types/RosettaTypeProvider.java
```diff
@@ -84,18 +85,23 @@ public RMetaAnnotatedType getRMetaAnnotatedType(AnnotationPathExpression expr) {
...
             }
+            
+        	RosettaType attributeType = a.getTypeCall().getType();
+        	if (attributeType instanceof Data data) {
+        		acc.addAll(getRMetaAttributesOfType(data));
+        	}
         }
         if (symbol instanceof Annotated ann) {
-            return getRMetaAttributes(ann.getAnnotations());
+            acc.addAll(getRMetaAttributes(ann.getAnnotations()));
```
SimonCockx, line 102: One more thing - if I'm correct, then an attribute is also an annotated. So for an attribute, I think these metaattributes will now be added twice? Both on line 93 and 102. Just checking.

PR #1108, opened 2025-09-24
Error when using then operation without implicit variable

davidalk:
This change will now throw an error on the Rune syntax when you use a `then` operation where the it's inline function has no implicit variables.

## Type of change


docs/rune-modelling-component.md
```diff
@@ -1160,6 +1160,42 @@ The Rune DSL provides a number of list operators that feature in usual programmi
 
 The following sections details the syntax and usage and these list operator features.
 
+#### Then
+
+The `then` keyword, when used within an expression, evaluates the expression on the left-hand side and passes its result as the input to the expression on the right-hand side.
+You can pass the entire left-hand result (for example, a list) to a function that expects it, as in the following example:
```
hugohills-regnosys, line 1166: you could add "for example a list, or empty"

docs/rune-modelling-component.md
```diff
@@ -1160,6 +1160,42 @@ The Rune DSL provides a number of list operators that feature in usual programmi
...
+For example, the following is incorrect:
+
+```Haskell
+Foo { isAllowable: False } filter isAllowable then "someResult"
+```
+
+If you intend to conditionally return "someResult" based on `isAllowable`, ensure that the right-hand side actually consumes the bound `item` by using `extract`, for example:
+
+```Haskell
+Foo { isAllowable: False } filter isAllowable then extract "someResult"
```
hugohills-regnosys, line 1196: I think we need some equivalent to this in Demo code model

docs/rune-modelling-component.md
davidalk: Done

rosetta-integration-tests/src/test/java/com/regnosys/rosetta/validation/ExpressionValidatorTest.java
```diff
@@ -22,6 +22,74 @@ public class ExpressionValidatorTest {
...
+    	RosettaExpression expr = modelService.toTestModel("""
+                type Foo:
+                     isAllowable boolean (1..1)
+               """).parseExpression("""
+               Foo { isAllowable: False } filter isAllowable then "someResult"
+               """);
+    	
+    	validationTestHelper.assertError(expr, INLINE_FUNCTION, null,
+                "The input item is not used in the `then` expression");
+    }
```
SimonCockx, line 92: Might be worth adding tests with nested `item` cases.

E.g., `"input" then ("hello" extract item)` should raise an error since the `"input"` is never used, even though `item` is present.

Similarly, `"input" then (Foo { isAllowable: False } extract isAllowable)` should throw an error, even though `isAllowable` refers to an attribute of the implicit variable, but not the right one.

docs/rune-modelling-component.md
```diff
@@ -1160,6 +1160,50 @@ The Rune DSL provides a number of list operators that feature in usual programmi
...
+There are situations where you might accidentally use `then` even though the right-hand side has no input parameter. 
+This is semantically incorrect because the left-hand result would be discarded. To prevent this, Rune raises a syntax error indicating that the input `item` is not used in the `then` expression. 
+For example, the following is incorrect:
+
+```Haskell
+    add ownersName: <"Filter lists to only include drivers with first and last names, then use 'map' to convert driving licences into list of names.">
+        drivingLicences
+            extract person
+            then filter firstName exists and surname exists
+            then firstName + " " + surname
```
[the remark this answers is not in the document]
davidalk: Updated

PR #1138, opened 2025-10-22
Fix issue to allow switching over types that exist in other namespaces

davidalk:
Fix issue to allow switching over types that exist in other namespaces

## Type of change


rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/expression/SwitchOperationTest.java
```diff
@@ -20,6 +25,56 @@ public class SwitchOperationTest {
     private RosettaTestModelService modelService;
     @Inject
     private JavaTypeUtil typeUtil;
+    @Inject
+    private FunctionGeneratorHelper functionGeneratorHelper;
+    @Inject
+    private CodeGeneratorTestHelper generatorTestHelper;
```
SimonCockx: Do you need to use these old utilities?

Why not `modelService`?

PR #1140, opened 2025-11-03
Function overriding preparation: refactoring

SimonCockx:

## Type of change

### DIFF
--- a/rune-ide/vscode/package-lock.json
+++ b/rune-ide/vscode/package-lock.json
@@ -91,31 +91,116 @@
         "typed-rest-client": "^1.8.4"
       }
     },
+    "node_modules/b4a": {
+      "version": "1.7.3",
+      "resolved": "https://registry.npmjs.org/b4a/-/b4a-1.7.3.tgz",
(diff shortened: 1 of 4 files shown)


rune-integration-tests/src/test/java/com/regnosys/rosetta/validation/NamesValidationTest.java
```diff
@@ -0,0 +1,89 @@
...
+            """);
+    }
+
+    @Test
+    void testDuplicateTypeInDifferentNamespaces() {
+        var models = modelHelper.parseRosetta("""
+                type Foo:
+                """, """
+                type Foo:
+                """);
```
davidalk: Shouldn't one of these be in a different namespace and the other import it? They will both be in the test namespace here right?

SimonCockx: Yes, this test is checking that types with the same qualified name (same namespace + same name) are disallowed, even if they are present in different files. Right now this is not checked - but it certainly results in runtime problems, because the Java code generator would overwrite once java file with the other. This change adds a check for it.

davidalk (approved):

PR #1207, opened 2026-04-29
Convert warnings to errors

### LINKED ISSUE #717: `regulatoryReference` should be `docReference`
`regulatoryReference` is exactly the same as `docReference`, so we should remove this feature.

### LINKED ISSUE #888: Cardinality warnings should be errors
Right now, most cardinality problems detected by the type system are marked as warnings. They should be marked as errors to prevent these problems from existing in Rune models.

Although this is a two-line change, this would be a backwards incompatible change, so we should plan this carefully.

### LINKED ISSUE #985: override keyword should be mandatory when overriding attributes
Right now, one can override an attribute by specifying an attribute with the same name as an attribute in a supertype. This behaviour has been deprecated in favour of explicit overrides using the `override` keyword. We should enforce this keyword in a next major version.

### LINKED ISSUE #991: Cardinality mismatch between rule reference and attribute should be an error
Right now they are a warning. They should become an error in next major version, since this is a backwards incompatible change.

### LINKED ISSUE #1198: Choice type path validation
Following the implementation of `switch` operation for `choice` type (https://github.com/finos/rune-dsl/pull/850), add validation to prohibit usage of choice types via path operators. 

Add warning or error when `choice` accessed via path, e.g. `asset -> basket`

Instead a `switch` should be used:
```
asset switch
  Basket then <item is now of type Basket>,
  Loan then <item is now of type Loan>,
  Security then <item is now of type Security>,
  default then <optional default case>
```

### LINKED ISSUE #1199: Choice type `only exists` validation
Choice types by design can only have one attribute populated, so using it with `only exists` operator should be prohibited as it will always return true.

For prod model versions (e.g. CDM 6.x.x), set as a warning, for dev model versions (e.g. CDM 7-dev), set as an error.

### LINKED ISSUE #1202: Warnings that should be errors
- `checkAnnotationSource` (duplicate types) in `RosettaSimpleValidator`
- Mandatory `[metadata key]` on types that are referenced.

davidalk:


rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/object/ModelObjectGeneratorTest.xtend
davidalk, line 1: I'll convert the xtend files in a follow up PR

rune-lang/src/main/java/com/regnosys/rosetta/validation/expression/AbstractExpressionValidator.java
SimonCockx, line 135: This method is now exactly the same as `isSingleCheckError`. These can now be de-duplicated. Lot's of methods to delete!

davidalk: done

SimonCockx (approved):

PR #1239, opened 2026-06-02
Fix formatter failure on non contiguous import blocks

davidalk:


rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/ResourceFormatterServiceTest.java
```diff
@@ -45,16 +45,26 @@ private void testFormatting(Collection<String> inputUrls, Collection<String> exp
...
-			formattedText.add(formattedContent); // Collect formatted content for assertions
+			formattedText.add(normalizeLineEndings(formattedContent)); // Collect formatted content for assertions
 		});
 
 		Assertions.assertIterableEquals(expectedText, formattedText);
 	}
 
+	private static String normalizeLineEndings(String text) {
+		return text.replace("\r\n", "\n");
+	}
```
davidalk: This is required because `ImportManagementService.toString(List<Import> imports)` adds a hard coded line ending which is different on Windows

hugohills-regnosys (approved):

rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/ResourceFormatterServiceTest.java
SimonCockx: Weird. Why wasn't this necessary before?

I would think that the expectations on Windows (due to git autocrlf option) would also contain \r\n line endings, so they would match up...

davidalk: Because before we had no tests here that triggered the `ImportManagementService` to add a new line. That service adds a new line appropriate for the OS.

SimonCockx (approved):

PR #1252, opened 2026-06-10
Reference choice types through an implied metadata key

davidalk:


rune-lang/src/main/java/com/regnosys/rosetta/Rosetta.xtext
```diff
@@ -72,7 +72,6 @@ Data:
 
 Choice:
 	'choice' RosettaNamed ':' RosettaDefinable?
-	Annotations*
```
SimonCockx, line 75: Removing `Annotations*` here drops support for *all* annotations on choices, not just `[metadata key]` — e.g. `[deprecated]` no longer parses, and the `supportDeprecatedAnnotationOnChoice` test was deleted along with it. Was that intentional?

If the goal is only to get rid of `[metadata key]` on choices, a targeted validation error on that annotation would be friendlier than a parse error on every annotation. If dropping all annotations is deliberate, it should be called out explicitly as a breaking change (any model with `[deprecated]` on a choice will fail to parse).

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/DeepPathUtilGenerator.xtend
```diff
@@ -62,68 +66,148 @@ class DeepPathUtilGenerator extends RObjectJavaClassGenerator<RDataType, JavaCla
...
+				typeUtil.STRING)
+		}
+		return leafKeyAccess(optionMapper, option.RMetaAnnotatedType.hasAttributeMeta, scope)
+	}
+
+	/**
+	 * Reads {@code getMeta().getExternalKey()} from a leaf option. When the option additionally carries
+	 * field-level metadata, its value is unwrapped via {@code getValue()} before reading the metadata.
+	 */
+	private def JavaExpression leafKeyAccess(JavaExpression optionMapper, boolean hasFieldMeta, JavaStatementScope scope) {
```
SimonCockx: This hand-rolls the `.map("getMeta", ...).map("getExternalKey", ...)` chain that `ExpressionGenerator.buildMapFunc` already generates for `-> key` on a regular keyed type — including re-hardcoding the `key` → `externalKey` mapping that is supposed to live in one place (`PojoPropertyUtil.toPojoPropertyName`). If that representation ever changes, we'd now have to update two places.

Could we extract the meta-read generation into a shared helper and reuse it here? As a bonus, that would keep this generator in the `JavaStatementBuilder` idiom instead of introducing a second style (raw string templates ending in `.get()`) within the same file.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/DeepPathUtilGenerator.xtend
```diff
@@ -62,68 +66,148 @@ class DeepPathUtilGenerator extends RObjectJavaClassGenerator<RDataType, JavaCla
...
+		val meta = scope.lambdaScope.createUniqueIdentifier("withMeta")
+		val metaFields = scope.lambdaScope.createUniqueIdentifier("metaFields")
+		if (hasFieldMeta) {
```
SimonCockx, line 195: I don't see a test exercising this branch — a choice option carrying attribute-level metadata, so the key is read through `getValue()` first. Could you add one to `DeepPathTest`?

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/DeepPathUtilGenerator.xtend
```diff
@@ -62,68 +66,148 @@ class DeepPathUtilGenerator extends RObjectJavaClassGenerator<RDataType, JavaCla
 				return false
 			])
 		])
+		val hasImpliedKey = choiceType.EObject instanceof Choice && (choiceType.EObject as Choice).buildRChoiceType.hasImpliedKey
```
SimonCockx: This `EObject instanceof Choice && (EObject as Choice).buildRChoiceType.hasImpliedKey` dance appears twice (here and in `streamObjects`, so it's also computed twice per type). The underlying gap is that `RChoiceType.asRDataType()` has no inverse, so code living in `RDataType`-land has to go via the `EObject`.

Two options to clean this up:
- add `Optional<RChoiceType> asChoiceType()` to `RDataType` (it already holds an `objectFactory`, so this is a one-liner), making this `choiceType.asChoiceType.map[hasImpliedKey].orElse(false)`; or
- add a `hasImpliedKey(RDataType)` overload to `DeepFeatureCallUtil`, next to `isEligibleForDeepFeatureCall`, that hides the choice check.

Either way, please extract it into a single helper so it's written (and evaluated) once.

rune-lang/src/main/java/com/regnosys/rosetta/types/RChoiceType.java
```diff
@@ -64,6 +63,43 @@ public List<RChoiceOption> getOwnOptions() {
...
+			if (optionRType instanceof RChoiceType) {
+				if (!((RChoiceType) optionRType).doHasImpliedKey(visited)) {
+					return false;
+				}
+			} else if (optionRType instanceof RDataType) {
+				RDataType leaf = (RDataType) optionRType;
+				// A leaf is keyed if it (or any of its supertypes) is annotated with [metadata key],
+				// mirroring the reference-target check in AttributeValidator.
+				boolean keyed = leaf.hasMetaAttribute("key")
+						|| leaf.getAllSuperTypes().stream().anyMatch(st -> st.hasMetaAttribute("key"));
```
SimonCockx: `getAllSuperTypes()` includes the type itself (see its Javadoc), so the `leaf.hasMetaAttribute("key") ||` clause on the previous line is redundant.

Also, this "keyed directly or via a supertype" idiom now exists here *and* in `AttributeValidator.checkAttribute` — worth extracting into something like `RDataType.hasInheritedMetaAttribute(String)` and using it in both places.

rune-lang/src/main/java/com/regnosys/rosetta/types/RChoiceType.java
```diff
@@ -64,6 +63,43 @@ public List<RChoiceOption> getOwnOptions() {
 		return ownOptions;
 	}
 	
+	/**
+	 * Returns true if every leaf option type of this choice (recursively through nested choices) has a [metadata key].
+	 * Returns false for an empty choice, an unresolved option, or any unkeyed leaf.
+	 */
+	public boolean hasImpliedKey() {
```
SimonCockx, line 71: This walks the full option tree (recursively through nested choices) and is invoked from scoping for every `->>` query on a choice. `RChoiceType` lazily caches its other derived state (`ownOptions`, `symbolId`, ...) — caching this boolean the same way would be a cheap win.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/ExpressionGenerator.xtend
```diff
@@ -783,7 +783,28 @@ class ExpressionGenerator extends RosettaExpressionSwitch<JavaStatementBuilder,
...
 	override protected caseDeepFeatureCall(RosettaDeepFeatureCall expr, Context context) {
-		return attributeCall(expr.receiver.javaCode(context.withExpected(MAPPER.wrapExtendsWithoutMeta(expr.receiver))), typeProvider.getRMetaAnnotatedType(expr.receiver), expr.feature.buildRAttribute, true, context.expectedType, context.scope)
+		val feature = expr.feature
+		if (feature instanceof RosettaMetaType) {
+			return deepMetaCall(expr.receiver.javaCode(context.withExpected(MAPPER.wrapExtendsWithoutMeta(expr.receiver))), typeProvider.getRMetaAnnotatedType(expr.receiver), feature, context.scope)
+		}
+		return attributeCall(expr.receiver.javaCode(context.withExpected(MAPPER.wrapExtendsWithoutMeta(expr.receiver))), typeProvider.getRMetaAnnotatedType(expr.receiver), (feature as Attribute).buildRAttribute, true, context.expectedType, context.scope)
+	}
+
+	def JavaStatementBuilder deepMetaCall(JavaStatementBuilder receiverCode, RMetaAnnotatedType receiverType, RosettaMetaType feature, JavaStatementScope scope) {
```
SimonCockx: A few things on this method:

- The receiver-handling preamble (choice → data view, lambda-param plumbing) duplicates the start of `attributeCall`. Could this be folded into the existing `metaCall`/`attributeCall` family rather than being a third variant?
- The result type is hardcoded to `String`; `metaCall` derives it from the feature via `typeProvider.getRTypeOfFeature(feature, null)` — better to do the same here.
- Beyond the type check, `feature` is ignored: any `RosettaMetaType` arriving here would generate a `metaChooseKey` call. Today scoping only ever surfaces `key` for deep feature calls, but that invariant is implicit and lives in a different file — deriving the result type from the feature and guarding on the feature name would make this robust if other meta types ever become deep-resolvable.

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
```diff
@@ -429,12 +431,18 @@ private IScope createExtendedFeatureScope(EObject receiver, RMetaAnnotatedType m
...
-		RType t = receiverType;
-		if (receiverType instanceof RChoiceType choice) {
-			t = choice.asRDataType();
+	private IScope createDeepFeatureScope(RType receiverType, EObject context) {
+		if (receiverType instanceof RChoiceType choiceType) {
+			RDataType dataView = choiceType.asRDataType();
+			IScope attributeScope = Scopes.scopeFor(Iterables.filter(Iterables.transform(deepFeatureCallUtil.findDeepFeatures(dataView), RAttribute::getEObject), Objects::nonNull));
+			if (choiceType.hasImpliedKey()) {
+				// Surface the `key` meta-feature so `someChoice ->> key` links when every leaf option is keyed.
+				List<RosettaFeature> keyMeta = ecoreUtil.getMetaDescriptions(List.of(new RMetaAttribute("key", null)), context);
```
SimonCockx: `new RMetaAttribute("key", null)` is a synthetic, half-constructed object (null type) used purely to smuggle a name into `getMetaDescriptions`, which immediately throws everything but the name away. An overload of `getMetaDescriptions` that takes names directly (e.g. `Set<String>`) would be more honest and avoid the dummy object.

SimonCockx (commented):
.

rune-lang/src/main/java/com/regnosys/rosetta/Rosetta.xtext
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/ExpressionGenerator.xtend
davidalk: Don't agree with the third point here as we don't have any other deep meta feature other than key yet, let's do it when we need it.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/DeepPathUtilGenerator.xtend
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/DeepPathUtilGenerator.xtend
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/DeepPathUtilGenerator.xtend
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/types/RChoiceType.java
davidalk: done

rune-lang/src/main/java/com/regnosys/rosetta/types/RChoiceType.java
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/ExpressionGenerator.xtend
davidalk: Done points 1 and 2

davidalk: I changed my mind, I agree with the last point and added a guard against it.

SimonCockx (approved):
.

Issue #938, opened 2025-02-27
Move to Google Guice 7


SimonCockx:
Google Guice 7 now relies on the `jakarta.inject` instead of the deprecated `javax.inject` namespace. To move over, we temporarily use Guice 6, which supports both packages and our generated code still emits `javax.inject` imports. With the next major version, we should move over to `jakarta.inject` and remove `javax.inject` entirely.

SimonCockx:
We should disallow non-jakarta inject imports by modifying our checkstyle config:
```
<!DOCTYPE module PUBLIC
  "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
  "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
<module name="Checker">
	<property name="localeCountry" value="UK"/>
	<property name="localeLanguage" value="en"/>
	<module name="TreeWalker">
		<!-- Forbid using Google Inject and the deprecated javax Inject in favor of jakarta Inject. -->
		<module name="IllegalImport">
			<property name="illegalClasses"
				value="com.google.inject.Inject, com.google.inject.name.Named, com.google.inject.Provider" />
			<property name="illegalPkgs"
				value="javax.inject" />
		</module>
	</module>
	<module name="BeforeExecutionExclusionFileFilter">
		<!-- Exclude any generated source folders -->
		<property name="fileNamePattern" value=".*-gen[\\/].*$" />
	</module>
</module>
```

PR #1256, opened 2026-06-15
Move configuration classes from rune-lang to rune-runtime

SimonCockx:
## Summary

**PR 1 of the "link XML config to a Rune function" stack** ([design](https://app.notion.com/p/37579b2a676b80088313c0bd33df7994)). Enabling work, no behaviour change.

Moves the `com.regnosys.rosetta.config[.file].*` configuration classes and the `ExperimentalFeature` enum from **rune-lang** into **rune-runtime** (package names unchanged), so an IDE/tool depending only on rune-runtime can read Rune configuration. `RosettaConfigExtension` and `ExperimentalFeatureService` stay in rune-lang (EMF-dependent).

## Details

- rune-runtime targets **Java 8** and mandates **javax.inject** (issue #938), so the moved classes are adapted: `List.of`/`Path.of` → Java 8 equivalents; `jakarta.inject` → `javax.inject`.
- The `RosettaConfiguration` provider binding in `RosettaRuntimeModule` is now wired explicitly through the Guice `Binder` (Xtext's `provide*` convention is typed to `jakarta.inject`, which rune-runtime forbids).
- Adds `jackson-dataformat-yaml` and `slf4j-api` to `rune-runtime/pom.xml`.

## Stack

This is the first PR in a dependency-ordered stack (PR1 → PR2 rename → PR3 readOnlyNamespaces → PR4 schema syntax → PR5 transform annotations → PR5b multi-config → PR6 hot-reload). Branches target `main`, so later PRs show cumulative diffs until the ones below merge; review/merge in order.

## Testing

`mvn install` of rune-runtime + rune-lang; `RosettaConfigurationTest` and `ExperimentalFeatureValidatorTest` pass (full injector + config + enum wiring).


rune-runtime/pom.xml
```diff
@@ -58,6 +58,14 @@
...
             <artifactId>jackson-datatype-jdk8</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-yaml</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
```
SimonCockx, line 68: These depenendencies moved from rune-lang. Are they still necessary there?

rune-lang/src/main/java/com/regnosys/rosetta/RosettaRuntimeModule.java
```diff
@@ -125,8 +125,10 @@ public Class<? extends IResourceValidator> bindIResourceValidator() {
 		return CachingResourceValidator.class;
 	}
 	
-	public Class<? extends Provider<? extends RosettaConfiguration>> provideRosettaConfiguration() {
-		return FileBasedRosettaConfigurationProvider.class;
+	// Bound explicitly via the Guice Binder because the provider now lives in rune-runtime,
+	// which mandates javax.inject (not the jakarta.inject convention used by Xtext's provide* methods).
+	public void configureRosettaConfiguration(Binder binder) {
+		binder.bind(RosettaConfiguration.class).toProvider(FileBasedRosettaConfigurationProvider.class);
```
SimonCockx, line 131: Instead of this change, would it be better to keep implementing both javax and jakarta Providers? What is the consequence for consumers? They won't be able to just directly use these classes as providers?

E.g., `public class FileBasedRosettaConfigurationProvider implements Provider<RosettaConfiguration>, jakarta.inject.Provider<RosettaConfiguration> {`

rune-runtime/pom.xml
SimonCockx: Good catch — fixed in the latest push:

- **`jackson-dataformat-yaml`**: only the config provider (`FileBasedRuneConfigurationProvider`, which builds the YAML `ObjectMapper`) used it. Since that moved to rune-runtime, rune-lang no longer references YAML anywhere (`grep` for `YAMLFactory`/`dataformat.yaml` in `rune-lang/src` returns nothing), so I removed the direct declaration from `rune-lang/pom.xml`. It now lives only in rune-runtime.
- **`slf4j-api`**: this one didn't actually "move" — rune-lang never declared it directly (it comes in transitively). I added it to rune-runtime because the moved `RuneConfigurationFileProvider` now logs the legacy-fallback warning, so runtime uses slf4j directly and should declare it.

rune-lang/src/main/java/com/regnosys/rosetta/RosettaRuntimeModule.java
SimonCockx: The reason I didn't keep the dual `javax`+`jakarta` interface is rune-runtime's own policy: `checkstyle-for-deprecated-guice.xml` forbids **both** `jakarta.inject` and `com.google.inject` across the whole module (it's the guard for #938). The classes could be dual while they lived in rune-lang, but rune-runtime mandates `javax.inject` only.

**Consequence for consumers — they can still use these classes directly as providers.** Guice 6 (our version) exposes `toProvider(Class<? extends javax.inject.Provider<? extends T>>)`, so a plain-Guice consumer binds them directly, e.g. `bind(RuneConfiguration.class).toProvider(FileBasedRuneConfigurationProvider.class)` — which is exactly what the explicit `Binder` line does. `javax.inject.Provider` is also the broadly-compatible interface.

The *only* thing that stopped working is Xtext's reflective `provide*` convention: Xtext 2.43 hardcodes it to `jakarta.inject.Provider`, so a `provideRuneConfiguration()` returning a `javax`-typed provider isn't recognised. That's framework-specific sugar, not general consumer usage — hence the one-line explicit binding (with a comment explaining why).

Going dual would re-introduce `jakarta.inject` into rune-runtime's API + classpath, contradicting the module's javax-only policy and tripping its checkstyle. It's doable, but only by carving out a suppression for the `config` package.

**My recommendation:** keep javax-only here for consistency with the module policy, and revisit dual holistically when #938 is resolved (runtime migrating to jakarta wholesale). But if you'd rather preserve the dual-provider ergonomics now, I'm happy to switch back to `implements Provider<…>, jakarta.inject.Provider<…>`, restore the `provideRuneConfiguration()` convention method, and scope a checkstyle suppression for `com.regnosys.rosetta.config`. Let me know which you prefer.

SimonCockx: Confirmed offline: we'll **keep javax-only** here, consistent with rune-runtime's #938 policy. The providers remain directly usable via plain Guice; only the one-line explicit `Binder` in rune-lang is needed. We can revisit a dual/jakarta migration holistically when #938 is addressed. No code change needed — leaving as-is.

JayasriR (approved):

PR #1257, opened 2026-06-15
Rename rosetta-config to rune-config

SimonCockx:
## Summary

**PR 2 of the "link XML config to a Rune function" stack** ([design](https://app.notion.com/p/37579b2a676b80088313c0bd33df7994)). Depends on #1256 (PR1).

Renames the configuration file `rosetta-config.yml` → `rune-config.yml` and the `Rosetta*Configuration` Java classes to `Rune*Configuration` (package `com.regnosys.rosetta.config[.file]` unchanged).

## Details

- Class renames: `RosettaConfiguration` → `RuneConfiguration`, `RosettaModelConfiguration` → `RuneModelConfiguration`, `RosettaGeneratorsConfiguration` → `RuneGeneratorsConfiguration`, `RosettaDependencyConfiguration` → `RuneDependencyConfiguration`, `RosettaConfigurationFileProvider` → `RuneConfigurationFileProvider`, the `Default*`/`FileBased*`/`*Mixin` providers, and the config `RosettaAttributeReference` → `RuneAttributeReference`. **Note:** the EMF/grammar type `RosettaAttributeReference` (`com.regnosys.rosetta.rosetta`, in `Rosetta.xcore`/`.xtext`) is a *different* class and is intentionally left untouched.
- `RuneConfigurationFileProvider` resolves `rune-config.yml` first and **falls back to the legacy `rosetta-config.yml` with a one-time deprecation warning**, for both classpath and file lookups.
- Importers/test helpers updated; binder method renamed to `configureRuneConfiguration`.
- Committed test resources renamed to `rune-config.yml`, **except** rune-ide's, kept as `rosetta-config.yml` to exercise the legacy fallback.

## Stack / diff note

Stacked on #1256 and targets `main`, so the diff currently includes PR1's changes; it will shrink to just the rename once #1256 merges. Review/merge in order.

## Testing

`RuneConfigurationTest`, `ExperimentalFeatureValidatorTest`, `FunctionExtensionValidatorTest`, `ModelMetaGeneratorFilteredNamespaceTest`, and the rune-ide formatter test (which now exercises the legacy fallback) all pass.

PR #1258, opened 2026-06-15
Add readOnlyNamespaces config and read-only-namespace CI check

SimonCockx:
## Summary

**PR 3 of the "link XML config to a Rune function" stack** ([design](https://app.notion.com/p/37579b2a676b80088313c0bd33df7994)). Orthogonal to the xml-config story but part of the same config overhaul. Depends on #1256, #1257 (both merged).

Lets a project mark namespaces as **read-only** — useful for namespaces that are managed externally (e.g. generated by importing an XSD or JSON schema) and so must not be hand-edited — and have CI reject manual changes to them.

- Adds `List<String> readOnlyNamespaces` to `RuneConfiguration` (empty default, getter, deserialized via the mixin). Pure data — no DSL behaviour.
- Adds a reusable workflow `verify-readonly-namespaces.yml` (`on: workflow_call`, owns its `fetch-depth: 0` checkout) and a self-contained, stdlib-only Python checker (`.github/scripts/check_readonly_namespaces.py`).
- Documents the feature under `website/docs/developers/read-only-namespaces.mdx`.

## How the check works

The read-only namespaces are read from the project's `rune-config.yml` `readOnlyNamespaces` list — a **single source of truth**, not a separate workflow input. For each changed `.rosetta` file the checker maps its namespace **before** (base ref) and **after** (HEAD) and fails if **either** is read-only:

- a file in a read-only namespace may not be modified, deleted, or moved *out* of it;
- a file may not be added to, or moved *into*, a read-only namespace.

It also fails if a configured pattern matches no namespace (stale/mistyped pattern). The namespace parser handles an optional `override namespace …` prefix.

## Making an intentional change

Adding the **`override-readonly-namespaces`** label (configurable via the `bypass-label` input) to a PR skips the check for that PR and records a notice — so an approved change can go through without removing the workflow or the namespace from `readOnlyNamespaces`. See the docs page for details.

## Caller usage

```yaml
jobs:
  readonly-namespaces:
    uses: finos/rune-dsl/.github/workflows/verify-readonly-namespaces.yml@main
    with:
      config-path: rune-config.yml
```

Inputs: `config-path` (required), `base-ref`, `root`, `checker-ref`, `bypass-label`.

## Stack / diff note

Stacked on the merged #1256/#1257 and targets `main`; later PRs (#1259–#1262) stack on this one. Review/merge in order.

## Testing

`RuneConfigurationTest` asserts `readOnlyNamespaces` parsing. The checker was exercised locally against an isolated git repo: `override` namespaces parse; modify/delete/move-out-of a read-only namespace → fail; add/move-into a read-only namespace → fail; edits in non-read-only namespaces → pass; stale pattern → fail; `readOnlyNamespaces` read from both block and inline YAML.

### DIFF
--- a/.github/scripts/check_readonly_namespaces.py
+++ b/.github/scripts/check_readonly_namespaces.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+"""Fail the build when a read-only Rune namespace is changed.
+
+The read-only namespaces are read from the ``readOnlyNamespaces`` list of a
+``rune-config.yml`` file, so they are configured in a single place. A change to a
+``.rosetta`` file is a violation when the file's namespace is read-only either:
+
+* **before** the change -- you may not modify, delete or move a file *out of* a
+  read-only namespace; or
+* **after** the change -- you may not add a file to, or move a file *into*, a
+  read-only namespace.
+
+The build also fails if a configured pattern matches no namespace in the
+repository, which catches stale or mistyped patterns.
+
+The checker is self-contained: it only uses the Python standard library and git.
+"""
+import argparse
+import fnmatch
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+# Matches `namespace foo.bar`, optionally preceded by the `override` keyword and optionally quoted.
+NAMESPACE_RE = re.compile(r'^\s*(?:override\s+)?namespace\s+"?([A-Za-z0-9_.]+)"?', re.MULTILINE)
+
+
+def git(*args):
+    return subprocess.run(["git", *args], capture_output=True, text=True, check=True).stdout
+
+
+def git_show(ref, path):
+    """Content of `path` at `ref`, or None if it does not exist there."""
+    result = subprocess.run(["git", "show", f"{ref}:{path}"], capture_output=True, text=True)
+    return result.stdout if result.returncode == 0 else None
+
+
+def namespace_of(text):
+    if text is None:
+        return None
+    match = NAMESPACE_RE.search(text)
+    return match.group(1) if match else None
+
+
+def read_readonly_namespaces(config_path):
+    """Read the `readOnlyNamespaces` list from a rune-config.yml file."""
+    config = Path(config_path)
+    if not config.is_file():
+        return []
+    patterns = []
+    in_block = False
+    for raw in config.read_text(encoding="utf-8").splitlines():
+        line = re.sub(r'\s+#.*$', '', raw).rstrip()  # strip trailing comments
+        if not line.strip():
+            continue
+        key = re.match(r'^readOnlyNamespaces\s*:\s*(.*)$', line)
+        if key:
+            inline = key.group(1).strip()
+            if inline.startswith("["):
+                patterns.extend(_clean(v) for v in inline.strip("[]").split(",") if _clean(v))
+                in_block = False
+            else:
+                in_block = True
+            continue
+        if in_block:
+            item = re.match(r'^\s*-\s*(.+?)\s*$', line)
+            if item:
+                value = _clean(item.group(1))
+                if value:
+                    patterns.append(value)
+            elif not line[:1].isspace():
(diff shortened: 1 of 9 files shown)


.github/scripts/check_readonly_namespaces.py
```diff
@@ -0,0 +1,120 @@
...
+"""
+import argparse
+import fnmatch
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+# Matches `namespace foo.bar.baz` (optionally quoted), the first such declaration in a file.
+NAMESPACE_RE = re.compile(r'^\s*namespace\s+"?([A-Za-z0-9_.]+)"?', re.MULTILINE)
```
SimonCockx: Note that a namespace can be preceded by other keywords such as "override". Is that supported?

.github/scripts/check_readonly_namespaces.py
```diff
@@ -0,0 +1,120 @@
...
+            continue
+        parts = line.split("\t")
+        status = parts[0][0]
+        path = parts[-1]  # for renames/copies this is the new path
+        if path.endswith(".rosetta"):
+            changed.append((status, path))
+    return changed
+
+
+def namespace_for_change(status, path, base):
```
SimonCockx: Suppose that I change the namespace from a protected namespace to an unprotected one. That should be prohibited, so I think the logic should always look at the original file before the change. So the logic:
- files that matched the pattern BEFORE may not have any changes
- files that did NOT match the pattern before (or which are newly added) may not change its namespace into a protected namespace.
Is that logic correctly implemented?

.github/workflows/verify-readonly-namespaces.yml
```diff
@@ -0,0 +1,86 @@
...
+#       uses: finos/rune-dsl/.github/workflows/verify-readonly-namespaces.yml@main
+#       with:
+#         protected-namespaces: |
+#           cdm.*
+#           com.rosetta.model.*
+#
+on:
+  workflow_call:
+    inputs:
+      protected-namespaces:
```
SimonCockx: This doesn't seem right. We now have to specify protected namespaces in two places. Instead this should be read directly from the rune-config.yml file, e.g., by specifying the path to that file.

.github/scripts/check_readonly_namespaces.py
SimonCockx: Good catch — it wasn't. Fixed: the namespace regex now allows an optional `override` prefix (`^\s*(?:override\s+)?namespace\s+"?([A-Za-z0-9_.]+)"?`), so `override namespace cdm.event` is recognised. Added a unit assertion for it.

.github/scripts/check_readonly_namespaces.py
SimonCockx: Agreed, and it wasn't fully covered. Reworked the logic to look at both states: for each changed `.rosetta` file I now compute its namespace **before** the change (from the base ref) and **after** (from HEAD), and fail if **either** is protected. So:
- a file whose namespace was protected before may not be modified, deleted, or moved out of the protected namespace (caught by the *before* check);
- a file may not be added to, or moved into, a protected namespace (caught by the *after* check).

Verified with scenarios: modify-in-protected → fail, move-out-of-protected → fail, add-to-protected → fail, edit/add in an unprotected namespace → pass.

.github/workflows/verify-readonly-namespaces.yml
SimonCockx: Agreed — removed the duplication. The workflow no longer takes a `protected-namespaces` list; it now takes a `config-path` input pointing at the project's `rune-config.yml`, and the checker reads the protected patterns directly from that file's `readOnlyNamespaces` list (single source of truth). Kept the checker stdlib-only with a small parser for the `readOnlyNamespaces` block so no extra dependency/pip step is needed.

JayasriR (approved):

PR #1259, opened 2026-06-15
Add schema syntax, SerializationFormat and transform annotations

SimonCockx:
## Summary

**PR 4 of the "link XML config to a Rune function" stack** ([design](https://app.notion.com/p/37579b2a676b80088313c0bd33df7994)). Depends on #1256, #1257, #1258. **Now also includes the transform annotations** (previously #1260) — schema and transform annotations are closely related, so they're a single PR.

## Schema, format and config

- **`SerializationFormat` enum** (`JSON`, `RUNE_JSON`, `XML`, `CSV`) in `basictypes.rosetta`, documented on the enum and each value. It maps onto the **canonical hand-written** `com.rosetta.model.lib.transform.SerializationFormat`: `basictypes` is excluded from code generation and `JavaTypeTranslator.caseEnumType` resolves the built-in to the hand-written type.
- **`schema <name> <format>`** — a new top-level construct (`RosettaRootElement` + `RosettaNamed`). `format` refers to a `SerializationFormat` value (scoped in `RosettaScopeProvider`, cached per basictypes resource). Schema names must be unique; `schema` is highlighted in `rosetta.tmLanguage.yaml`.
- **`serializationConfig`** — `RuneSerializationConfiguration {id, configPath}` POJO and a `serializationConfig` list on `RuneConfiguration` (`findSerializationConfigById`).

## Transform annotations

- **`TransformKind`** + **`TransformAnnotation`** — `[ingest <fmt-or-schema>]`, `[projection …]`, `[enrich]`. `ref` resolves to a `Schema` or a `SerializationFormat` value via a **`SchemaOrFormat`** marker interface. `ingest`/`enrich`/`projection` are added to `ValidID` (contextual keywords); the old annotation declarations are removed from `annotations.rosetta`.
- **`TransformAnnotationHelper`** resolves a transform annotation to `{schema id?, format, configPath?}`. Validation: one transform per function; single input; `enrich` carries no format; `ingest`/`projection` require a format; CSV must be tabular.
- **`FunctionGenerator`** emits `@Ingest`/`@Projection`/`@Enrich` (`com.rosetta.model.lib.transform`) on the generated function class.

## Stack / diff note

Stacked on #1256–#1258 (PR1/PR2 merged) and targets `main`; #1261/#1262 stack on this. Review/merge in order.

## Testing

`SchemaValidationTest`, `TransformAnnotationGeneratorTest`, plus updated `FunctionValidatorTest`/`RosettaValidatorTest`/`RuneConfigurationTest`. Formatter/parser/serializer regression subsets unaffected.


rune-integration-tests/src/test/java/com/regnosys/rosetta/config/RuneConfigurationTest.java
```diff
@@ -40,7 +40,15 @@ public Class<? extends RuneConfigurationFileProvider> bindRuneConfigurationFileP
 		assertFalse(filter.test("foo"));
 		assertTrue(filter.test("abc.def"));
 		assertFalse(filter.test("abc.def.sub"));
-		
+
+		assertEquals(java.util.List.of("com.rosetta.model.*", "abc.def"), config.getReadOnlyNamespaces());
+
+		assertEquals(2, config.getSerializationConfig().size());
+		assertEquals("xml-config/fixml-xml-config.json",
```
SimonCockx: Please do not reference existing schema's like fixml.

rune-integration-tests/src/test/java/com/regnosys/rosetta/validation/SchemaParsingTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.eclipse.xtext.testing.InjectWith;
+import org.eclipse.xtext.testing.extensions.InjectionExtension;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import com.regnosys.rosetta.tests.RosettaTestInjectorProvider;
+
+@ExtendWith(InjectionExtension.class)
+@InjectWith(RosettaTestInjectorProvider.class)
+public class SchemaParsingTest extends AbstractValidatorTest {
```
SimonCockx: Consistent with other validation tests - should this instead be "SchemaValidationTest"?

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/types/JavaTypeUtil.java
```diff
@@ -74,6 +74,7 @@ public class JavaTypeUtil {
 	public final JavaClass<com.rosetta.model.lib.records.Date> DATE = JavaClass.from(com.rosetta.model.lib.records.Date.class);
 	public final JavaClass<LocalDateTime> LOCAL_DATE_TIME = JavaClass.from(LocalDateTime.class);
 	public final JavaClass<ZonedDateTime> ZONED_DATE_TIME = JavaClass.from(ZonedDateTime.class);
+	public final JavaClass<com.rosetta.model.lib.transform.SerializationFormat> SERIALIZATION_FORMAT = JavaClass.from(com.rosetta.model.lib.transform.SerializationFormat.class);
```
SimonCockx: Please import it instead of a FQN.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/types/JavaTypeTranslator.java
```diff
@@ -309,8 +309,17 @@ public JavaType toJavaType(RMetaAnnotatedType type) {
 		return javaType;
 	}
 	private JavaType toJavaType(RType type) {
+		// The built-in SerializationFormat enum (declared in basictypes.rosetta) is not generated;
+		// it maps onto the canonical hand-written enum com.rosetta.model.lib.transform.SerializationFormat.
+		if (type instanceof REnumType && isBuiltInSerializationFormat((REnumType) type)) {
```
SimonCockx: Instead of this instance check, can it not just be put inside `caseEnumType`?

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
```diff
@@ -107,6 +107,16 @@ public IScope getScope(EObject context, EReference reference) {
...
+			} else if (reference.equals(SCHEMA__FORMAT)) {
+				// A schema's format refers to a value of the built-in SerializationFormat enum.
+				RosettaModel basicTypes = builtinsService.getBasicTypesModel(context.eResource().getResourceSet());
+				return basicTypes.getElements().stream()
+						.filter(RosettaEnumeration.class::isInstance)
+						.map(RosettaEnumeration.class::cast)
+						.filter(e -> "SerializationFormat".equals(e.getName()))
+						.findFirst()
+						.map(e -> (IScope) Scopes.scopeFor(e.getEnumValues()))
+						.orElse(IScope.NULLSCOPE);
```
SimonCockx: I wonder, in terms of efficiency, how well this fairs. This will need to run for each SCHEMA__FORMAT resolution... Can it be improved without loss of correctness?

rune-runtime/src/main/java/com/rosetta/model/lib/transform/SerializationFormat.java
```diff
@@ -0,0 +1,21 @@
...
+ * <p>
+ * This is the canonical Java representation of the {@code SerializationFormat} enum declared
+ * in {@code basictypes.rosetta}: the Rune code generator maps the model enum onto this type
+ * rather than generating a duplicate, so there is a single shared enum across the DSL, the
+ * generated annotations and the runtime.
+ */
+public enum SerializationFormat {
+    /** Plain JSON, mapping directly onto the structure of the Rune type. */
+    JSON,
+    /** The Rune-specific JSON representation, which preserves Rune metadata such as references and scheme information. */
```
SimonCockx: This comment is wrong. The plain JSON mapping also preserves this information. It is just a different standard. I think there might be a GitHub issue on this.

rune-runtime/src/main/resources/model/basictypes.rosetta
```diff
@@ -52,3 +52,10 @@ recordType zonedDateTime
 
 typeAlias calculation:
 	string
+
+enum SerializationFormat: <"The wire format used to (de)serialize the input or output of a transform function (e.g. an ingestion or projection).">
+	JSON <"Plain JSON, mapping directly onto the structure of the Rune type.">
+	RUNE_JSON <"The Rune-specific JSON representation, which preserves Rune metadata such as references and scheme information.">
```
SimonCockx: Idem dito

rune-integration-tests/src/test/java/com/regnosys/rosetta/config/RuneConfigurationTest.java
SimonCockx: Done — replaced the `fixml` references with a generic `myXmlSchema` (and the dependency-config one with `myOtherSchema`) across the tests and fixtures.

rune-integration-tests/src/test/java/com/regnosys/rosetta/validation/SchemaParsingTest.java
SimonCockx: Good point — renamed to `SchemaValidationTest`.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/types/JavaTypeUtil.java
SimonCockx: Resolved by removing the constant entirely — see the next comment; the mapping moved into `caseEnumType`, so the FQN/constant in `JavaTypeUtil` is no longer needed.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/types/JavaTypeTranslator.java
SimonCockx: Yes — moved it into `caseEnumType`. The reason it wasn't there originally is that `caseEnumType` returns `RJavaEnum` (consumed by `EnumGenerator`), whereas the SerializationFormat mapping needs to point at the hand-written runtime type. I addressed that by adding an `RJavaEnum(REnumType, JavaPackageName)` constructor and returning `new RJavaEnum(type, <transform package>)` for the built-in — so the logic now lives in `caseEnumType` and the separate instance-check + `JavaTypeUtil` constant are gone.

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
SimonCockx: Agreed. The lookup now caches the SerializationFormat enum via Xtext's `IResourceScopeCache`, keyed by the basictypes resource, so it's computed once rather than per `SCHEMA__FORMAT`/transform-ref resolution. (`getBasicTypesModel` itself is just a cheap resource-map lookup.)

rune-runtime/src/main/java/com/rosetta/model/lib/transform/SerializationFormat.java
SimonCockx: Fixed — reworded so neither value implies the other loses metadata: JSON is "following a standard JSON representation of the model" and RUNE_JSON is "following the Rune-specific JSON standard". They're two different JSON standards.

rune-runtime/src/main/resources/model/basictypes.rosetta
SimonCockx: Done — same rewording applied to the enum docs in basictypes.rosetta.

rune-ide/rosetta.tmLanguage.yaml
```diff
@@ -567,12 +568,24 @@ repository:
...
 
+      transformAnnotationBody:
+        name: meta.annotated.transform.rosetta
+        begin: '{{wordStart}}(ingest|enrich|projection){{wordEnd}}'
+        beginCaptures:
+          1: { name: keyword.other.transform.rosetta }
+        end: (?=\])|{{sectionEnd}}
+        patterns:
+        - include: '#comment'
+        - name: variable.other.enummember.rosetta
```
SimonCockx: This is not necessarily an enum member though, right? It can either be an enummember or a schema. Perhaps this should be a generic fallback and then the semantic token provider should provide it with the correct one?

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/function/TransformAnnotationGeneratorTest.java
```diff
@@ -0,0 +1,74 @@
...
+					inputs:
+						input Foo (1..1)
+					output:
+						result Foo (1..1)
+				""");
+
+		String all = String.join("\n", code.values());
+
+		assertTrue(all.contains("@Ingest(format = SerializationFormat.XML)"),
+				"bare-format ingest should emit @Ingest with just the format");
```
SimonCockx, line 67: Can this be made more specific? E.g., "this particular class file contains this string".

rune-lang/model/RosettaSimple.xcore
```diff
@@ -54,6 +55,21 @@ class AnnotationRef {
 	contains AnnotationQualifier[] qualifiers
 }
 
+enum TransformKind {
+	ingest
+	enrich
+	projection
```
SimonCockx: In terms of casing convention, what do other xcore enums use? Is this consistent?

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/function/FunctionGenerator.xtend
```diff
@@ -134,6 +145,21 @@ class FunctionGenerator extends RObjectJavaClassGenerator<RFunction, RGeneratedJ
...
+			return
+		}
+		val format = transformAnnotationHelper.getFormat(transform).orElse(null)
+		if (format === null) {
+			return // an ingest/projection without a resolvable format is reported by validation
+		}
+		val id = transformAnnotationHelper.getSchemaId(transform).orElse(null)
+		val configPath = transformAnnotationHelper.getConfigPath(transform).orElse(null)
+		val StringConcatenationClient attributes = '''«IF id !== null»id = "«id»", «ENDIF»format = «SerializationFormat».«format»«IF configPath !== null», configPath = "«configPath»"«ENDIF»'''
+		annotations.put(if (transform.kind === TransformKind.INGEST) Ingest else Projection, attributes)
```
SimonCockx: I wonder: would it be cleaner to put this mapping from transform kind to Java class inside the JavaTypeTranslator?

SimonCockx: I think enrich, ingest and projection can then be handled consistently without any case distinction here.

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
```diff
@@ -361,6 +371,21 @@ protected ImportNormalizer doCreateImportNormalizer(QualifiedName importedNamesp
...
+		RosettaEnumeration serializationFormat = resourceScopeCache.get(SERIALIZATION_FORMAT_ENUM_CACHE_KEY,
+				basicTypes.eResource(), () -> {
+					for (var element : basicTypes.getElements()) {
+						if (element instanceof RosettaEnumeration e && "SerializationFormat".equals(e.getName())) {
+							return e;
+						}
+					}
+					return null;
+				});
+		return serializationFormat == null ? java.util.Collections.emptyList() : serializationFormat.getEnumValues();
```
SimonCockx: FQN: please fix.

rune-lang/src/main/java/com/regnosys/rosetta/utils/TransformAnnotationHelper.java
```diff
@@ -0,0 +1,62 @@
...
+		if (ref instanceof RosettaEnumValue formatValue) {
+			return Optional.ofNullable(formatValue.getName());
+		}
+		if (ref instanceof Schema schema && schema.getFormat() != null) {
+			return Optional.ofNullable(schema.getFormat().getName());
+		}
+		return Optional.empty();
+	}
+
+	public boolean hasFormat(TransformAnnotation annotation, String formatName) {
```
SimonCockx: Instead of a String based formatName, couldn't we use the SerializationFormat enum?

rune-ide/rosetta.tmLanguage.yaml
SimonCockx: Good point — changed it to a generic `variable.other.rosetta` fallback (with a comment noting the ref can be either a serialization-format enum value or a schema name). Semantic highlighting can refine it later via the token provider.

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/function/TransformAnnotationGeneratorTest.java
SimonCockx: Done — the test now asserts against the specific generated class file for each function (via a `classCode(code, simpleName)` helper) instead of the concatenation of all generated files.

rune-lang/model/RosettaSimple.xcore
SimonCockx: Checked — `CardinalityModifier` uses the same `UPPER as "lower"` form (`NONE as "none"`, `ALL as "all"`, …), so `INGEST as "ingest"` is consistent with the existing convention for enums whose literals map to keywords.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/function/FunctionGenerator.xtend
SimonCockx: Done — added `JavaTypeTranslator.toTransformAnnotationClass(TransformKind)` which maps each kind to its annotation class (`Ingest`/`Projection`/`Enrich`), and the generator now uses it.

SimonCockx: Agreed — `addTransformAnnotation` no longer has any case distinction. It maps the kind to the annotation class via the translator and emits the optional `id`/`format`/`configPath` attributes uniformly for enrich/ingest/projection.

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
SimonCockx: Fixed — using the imported `Collections.emptyList()` now instead of the FQN.

rune-lang/src/main/java/com/regnosys/rosetta/utils/TransformAnnotationHelper.java
SimonCockx: Done — `getFormat` now returns `Optional<SerializationFormat>` and `hasFormat(annotation, SerializationFormat)` takes the enum, so callers (e.g. `FunctionValidator`, the generator) work with the enum directly. The only string→enum conversion is a private helper that resolves the model enum-value name to the runtime `SerializationFormat`.

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/function/TransformAnnotationGeneratorTest.java
```diff
@@ -0,0 +1,80 @@
...
+				"enrich should emit @Enrich");
+	}
+
+	private static String classCode(Map<String, String> code, String simpleName) {
+		return code.entrySet().stream()
+				.filter(e -> e.getKey().endsWith("." + simpleName))
+				.map(Map.Entry::getValue)
+				.findFirst()
+				.orElseThrow(() -> new AssertionError("No generated class found for " + simpleName));
+	}
```
SimonCockx: There is actually already an existing utility for this. See JavaTestModel getJavaCode.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/function/FunctionGenerator.xtend
```diff
@@ -134,6 +142,18 @@ class FunctionGenerator extends RObjectJavaClassGenerator<RFunction, RGeneratedJ
 		rFunction.classBody(javaFunctionClass, isStatic, overridesEvaluate, dependencies, functionInterfaces, annotations, classScope)
 	}
 	
+	private def void addTransformAnnotation(TransformAnnotation transform, Map<Class<?>, StringConcatenationClient> annotations) {
+		val annotationClass = transform.kind.toTransformAnnotationClass
+		val format = transformAnnotationHelper.getFormat(transform).orElse(null)
+		if (annotationClass !== Enrich && format === null) {
```
SimonCockx: Well, if it is handled by validation. Why check for Enrich at all? It is elegant if there is no case distinction at all at this point.

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/function/TransformAnnotationGeneratorTest.java
SimonCockx: Good catch — switched the test to `RosettaTestModelService.toJavaTestModel(...)` and assert against `model.getFunctionJavaSource("IngestBareFormat")` etc., so each assertion targets that specific function class file. Removed the ad-hoc `classCode` helper.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/function/FunctionGenerator.xtend
SimonCockx: Agreed — removed the `Enrich` check entirely. `addTransformAnnotation` now has no case distinction at all: it maps the kind to the annotation class and emits whichever of `id`/`format`/`configPath` are present. For `enrich` all three are absent, so it renders as a bare `@Enrich`. (The `Enrich` import is gone too.)

rune-ide/rosetta.tmLanguage.yaml
SimonCockx: Follow-up: the semantic token provider now refines this. `RosettaSemanticTokensService.markTransformAnnotationRef` emits `enumMember` when the ref resolves to a serialization-format enum value and a new `schema` token type when it resolves to a `Schema` (the schema declaration and its format reference are marked too). So the TextMate `variable.other` is just the static fallback and the semantic layer provides the precise token.

SimonCockx: Correction to the above: I dropped the extra marking of the `schema` *declaration* — it is already fully highlighted by the TextMate grammar (`entity.name.schema.rosetta`, and the format value as `enumMember`), so re-marking it semantically was redundant. The semantic provider now only marks the ambiguous *ref*: `enumMember` for a format value, and a new `schema` token type for a schema reference. The `schema` token is registered in the VS Code extension (`rune-ide/vscode/package.json`) and scope-mapped to `entity.name.schema.rosetta`, so a reference renders the same as the declaration.

rune-lang/src/main/java/com/regnosys/rosetta/utils/TransformAnnotationHelper.java
```diff
@@ -0,0 +1,75 @@
...
+ * and the optional config file path (looked up by schema id in the Rune configuration).
+ */
+@Singleton
+public class TransformAnnotationHelper {
+	@Inject
+	private RuneConfiguration config;
+
+	public List<TransformAnnotation> getTransformAnnotations(Function function) {
+		return function.getTransform();
+	}
```
SimonCockx: Get rid of this.

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
```diff
@@ -93,8 +93,12 @@ public class RosettaScopeProvider extends ImportedNamespaceAwareLocalScopeProvid
 	private DeepFeatureCallUtil deepFeatureCallUtil;
 	@Inject
 	private RObjectFactory rObjectFactory;
-	@Inject 
+	@Inject
 	private RosettaBuiltinsService builtinsService;
+	@Inject
+	private org.eclipse.xtext.util.IResourceScopeCache resourceScopeCache;
```
SimonCockx: FQN. Please fix.

rune-lang/src/main/java/com/regnosys/rosetta/utils/TransformAnnotationHelper.java
SimonCockx: Done — removed the redundant `getTransformAnnotations` passthrough (it only returned `function.getTransform()` and had no callers), along with its now-unused `List` import.

rune-lang/src/main/java/com/regnosys/rosetta/scoping/RosettaScopeProvider.java
SimonCockx: Fixed — imported `org.eclipse.xtext.util.IResourceScopeCache` and used the simple name instead of the FQN.

rune-runtime/src/main/resources/model/basictypes.rosetta
```diff
@@ -52,3 +52,11 @@ recordType zonedDateTime
 
 typeAlias calculation:
 	string
+
+enum SerializationFormat: <"The wire format used to (de)serialize the input or output of a transform function (e.g. an ingestion or projection).">
+	JSON <"JSON, following a standard JSON representation of the model.">
+	RUNE_JSON <"JSON, following the Rune-specific JSON standard.">
+	XML <"XML, typically configured by an associated XML configuration file that maps the schema onto the Rune type.">
+	CSV <"Comma-separated values. Only valid for tabular types, i.e. types whose attributes are all single-cardinality basic types.">
+	CSV_LABELLED <"Comma-separated values with a header row of human-readable labels sourced from a label provider. Only valid for tabular types.">
```
SimonCockx: Replace with "Comma-separated values with a header row of human-readable labels."

SimonCockx: Done — shortened the doc to "Comma-separated values with a header row of human-readable labels." (applied to both basictypes.rosetta and the SerializationFormat enum).

rune-lang/src/main/java/com/regnosys/rosetta/validation/names/RosettaNamesAreUniqueValidationHelper.java
```diff
@@ -121,7 +123,22 @@ protected void doCheckUniqueIn(IEObjectDescription description, Context context,
...
+        return !(isBuiltin(description) && isBuiltin(candidate));
+    }
+
+    private boolean isBuiltin(IEObjectDescription description) {
+        URI uri = description.getEObjectURI();
+        if (uri == null) {
+            return false;
+        }
+        String resourceName = uri.trimFragment().lastSegment();
+        return "basictypes.rosetta".equals(resourceName) || "annotations.rosetta".equals(resourceName);
```
SimonCockx, line 148: Not sure about this change.

JayasriR (approved):

PR #1293, opened 2026-06-21
Convert RuleGenerator, ReportGenerator and FunctionGenerator to Java and migrate them to the fluent API

SimonCockx:
Continues the Xtend → Java migration of the code generators, moving them **onto the fluent `CodeRenderer`/`CodeWriter` API**. Migrates the report generators (`RuleGenerator`, `ReportGenerator`) and the big one, `FunctionGenerator`.

## What changed

- **`RuleGenerator`, `ReportGenerator`, `FunctionGenerator`**: converted `.xtend` → `.java` and moved onto `FluentRObjectJavaClassGenerator`; `generateClass`/`rBuildClass` now return a `CodeRenderer`.
- **`FunctionGenerator`** is rendered with the fluent `CodeWriter` API (`out.writeln`/`out.indented`/…) instead of Xtend `'''…'''` templates. Its annotation map and the report annotations are now `CodeRenderer`s.
- Because `FunctionGenerator.rBuildClass` returns a `CodeRenderer`, `RuleGenerator`/`ReportGenerator` no longer wrap its result — the call-site bridging is gone.

## Bridges

The migration is incremental, so two small bridges — both extracted out of `JavaExpression`, counterparts of each other — let fluent generators interoperate with code that still produces Xtend `StringConcatenationClient`s:

- **`CodeWriterTargetStringConcatenation`** — a streaming bridge that forwards each appended object to the `CodeWriter`, so a `RecordingCodeWriter` registers imports and **defers identifier resolution** (keeping the scope open). `FunctionGenerator` uses it to embed the few fragments still generated as templates by not-yet-migrated code (method parameter lists, the nested class declaration, alias call arguments).
- **`TargetStringConcatenationCodeWriter`** — the reverse bridge: it exposes a `CodeWriter` that appends into the legacy Xtend template machinery, so a value produced by the fluent API can still be embedded in a not-yet-migrated Xtend template (e.g. `RGeneratedJavaClass#asClassDeclaration`, used by generators that haven't moved to the fluent API yet).

Both are to be removed once all generators use the fluent API.

**`FluentImportManager`** (new) is the fluent counterpart of `ImportManagerExtension`: given a class body as a `CodeRenderer`, it assembles the full file (package, resolved imports, body). `FluentJavaClassGenerator` injects it rather than exposing its own file-assembly logic as a method on the generator; `FunctionGeneratorHelper`, which needs to drive single-class generation directly to assert on one function's output, injects it the same way the not-yet-migrated tests (`TypeCoercionTest`, `ExpressionGeneratorTest`) inject `ImportManagerExtension`.

## Output formatting

The fluent `CodeWriter` defaults to four-space indentation and emits no trailing whitespace, so generated function/rule code is reformatted accordingly (tabs → spaces, no trailing whitespace, conventional Javadoc). The affected expectations are updated; **behaviour is unchanged** — all generated code still compiles and runs.

## Testing

Full multi-module build passes (`mvn install`), 0 Checkstyle violations. The full `rune-integration-tests` suite passes (**1537 tests, 0 failures**), including the byte-comparison generation regression tests and the 147 `FunctionGeneratorTest` cases (which compile and execute the generated functions).


rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/function/FunctionGeneratorHelper.java
```diff
@@ -106,8 +103,8 @@ public void assertToGeneratedFunctionWithOperations(CharSequence actualModel, Ch
 	private void generate(RFunction func) {
 		RGeneratedJavaClass<? extends RosettaFunction> typeRepresentation = generator.createTypeRepresentation(func);
 		JavaClassScope classScope = JavaClassScope.createAndRegisterIdentifier(typeRepresentation);
-		StringConcatenationClient classCode = generator.generateClass(func, typeRepresentation, "test", classScope);
-		String javaFileCode = importManager.buildClass(typeRepresentation.getPackageName(), classCode, classScope.getFileScope());
+		CodeRenderer classCode = generator.generateClass(func, typeRepresentation, "test", classScope);
+		String javaFileCode = generator.buildClass(typeRepresentation.getPackageName(), classCode, classScope.getFileScope());
```
SimonCockx: How come this pattern changes? E.g., I would have thought this would just use a similar fluent API service.

davidalk (approved):

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/function/FunctionGeneratorHelper.java
davidalk: Added `FluentImportManager` in 7aa66d46, the fluent counterpart of `ImportManagerExtension`. `FluentJavaClassGenerator` no longer declares a `buildClass` method at all - it now injects `FluentImportManager` and delegates to it, the same way `XtendJavaClassGenerator` injects `ImportManagerExtension`. `FluentImportManager#buildClass` is `public`, matching `ImportManagerExtension#buildClass`.

Why this was needed: `buildClass` had been made `public` on `FluentJavaClassGenerator` purely so `FunctionGeneratorHelper` could call it - it's declared in package `generator.java`, while `FunctionGeneratorHelper` is in `generator.java.function`, a different package, and not a subclass, so it wasn't reachable via `protected` either way. `FunctionGeneratorHelper` now injects `FluentImportManager` directly instead of reaching through the generator instance.

JayasriR (approved):

PR #1301, opened 2026-06-26
[10.x.x] - Preserve alias identity in `as` and `switch` operations 

### LINKED ISSUE #1300: Type aliases as choice options mis-navigate in `as`/`switch` (aliases stripped in findChoiceOptionPath)
Type aliases are accepted as choice options by the grammar and validators, but `as`/`switch` navigation strips the alias and matches on the underlying type, so options sharing an underlying type resolve to the wrong field.

This model parses, validates, and the `Identifier` type generates and compiles cleanly:

```rune
typeAlias ISIN:  string
typeAlias CUSIP: string

choice Identifier:
    ISIN
    CUSIP
```

The problem is at runtime navigation. `navigateToChoiceOption` (`ExpressionGenerator.xtend:1210`) strips the alias from the goal type, and `findChoiceOptionPath` (`TypeSystem.java:366-385`) strips the alias from each option type before comparing. Both `ISIN` and `CUSIP` strip to `string`, so the lookup matches the first option with that underlying type — `as CUSIP` and the corresponding `switch` case silently resolve to the `ISIN` attribute. No exception is thrown; the value is just read from the wrong field.

Note this only manifests when two or more options share the same underlying type; a single alias option happens to navigate correctly.

**Why this happens:** there is a mismatch between what the syntax permits (type aliases as choice options) and the generator, which discards alias identity when computing the option path.

**Possible resolutions** (not recommending one):
- (a) Add a validation error rejecting type aliases as choice options, leaving the generator as-is.
- (b) Make `findChoiceOptionPath` distinguish options that share an underlying type. Note this is more subtle than "stop stripping": the strip in `findChoiceOptionPath` does double duty.
    - The **leaf match** (`target.equals(stripFromTypeAliases(optionType))`) is where the bug is — `RAliasType.equals` keys on the type function, so `ISIN` and `CUSIP` are distinct *before* stripping and identical *after*. This comparison should be alias-sensitive (compare the unstripped option type, or match on the `ChoiceOption` itself, since the `as`/`switch` operand already references the specific option).
    - The **descend-into-nested-choice** check (`optionType instanceof RChoiceType`) still needs the strip: an alias of a choice (`typeAlias X: SomeChoice`) is not `instanceof RChoiceType` until stripped, so removing the strip here would break navigation into a nested choice referenced through an alias.

    The goal type stripped in `navigateToChoiceOption` (`ExpressionGenerator.xtend:1210`) must likewise stop being stripped. No Java-type preservation is needed: aliases over basic types are already erased to their underlying Java type (`JavaTypeTranslator.caseAliasType`), and the generated choice distinguishes its options by attribute name (`getISIN()`/`getCUSIP()`), so a correct lookup already emits the right getter.

davidalk:
This PR makes type aliases over choices and choice options behave consistently with the types they wrap. While fixing the navigation bug in #1300, several related defects surfaced where an alias was not treated the same as its underlying type; each is addressed below.

## 1. `as` / `switch` navigate to the wrong field when options share an underlying type (#1300)

When two choice options aliased the same base type (e.g. `ISIN` and `CUSIP` both over `string`), `as CUSIP` silently resolved to the `ISIN` field — the alias was stripped before matching, so both options looked identical.

**Fix** (`TypeSystem.findChoiceOptionPath`, `ExpressionGenerator.navigateToChoiceOption`): the leaf match is now alias-sensitive — it compares the *unstripped* option type, so options are distinguished by alias identity. The strip is kept only for the descend-into-nested-choice check, where it is still needed to see through an alias of a choice. `navigateToChoiceOption` correspondingly stops stripping the goal type.

## 2. Feature calls on an alias of a choice did not resolve or generate

For `typeAlias ChoiceAlias: SomeChoice`, a receiver typed as `ChoiceAlias` was an `RAliasType` rather than an `RChoiceType`, so a feature call such as `x -> OptionA` failed to resolve its feature and could not be generated.

**Fix**:
- `RosettaEcoreUtil.allFeaturesExcludingEnumValues` — resolves features through an alias to its referent.
- `ExpressionGenerator.choiceFeatureCall` — strips an alias receiver to the underlying choice/data type before generating the call.

## 3. Deep feature calls (`->>`) on an alias of a choice failed

Deep-feature scoping and code generation only recognised a direct `RChoiceType`, so `->>` through an alias of a choice did not resolve and the required `DeepPathUtil` dependency was not collected.

**Fix**:
- `RosettaScopeProvider.createDeepFeatureScope` — strips the receiver type before the choice check.
- `JavaDependencyProvider` — strips before the choice check so the `DeepPathUtil` dependency is collected.

## 4. Nested options behind an alias of a choice were invisible

`RChoiceType.getAllOptions` did not recurse into an option whose type was an alias of a nested choice, so those nested options were dropped. This underpins both `as`/`switch` navigation and overlap validation.

**Fix** (`RChoiceType.doGetAllOptions`): strips aliases when collecting nested options.

## 5. Choice overlap validation did not see through an alias of a nested choice

An option already included via an alias of a nested choice was not reported as overlapping, so invalid choices passed validation.

**Fix** (`ChoiceValidator.checkChoiceOptionsDoNotOverlap`): strips the option type before the choice check so nested options reached through an alias are included in the overlap comparison.

## 6. Subtyping treated aliases inconsistently

An alias was unwrapped when it appeared as the subtype, but the data-type branch took priority over the right-hand alias branch — so `alias(X) <: data` succeeded while `data <: alias(X)` failed.

**Fix** (`SubtypeRelation.isSubtypeOf`): the alias branches are checked before the data-type branch, so an alias is treated transparently in both directions.

## Tests

- `AsOperationTest` — `as` against two alias options sharing a base type (string, enum, data type) navigates to the correct field and yields empty for the other; plus alias-as-leaf and alias-as-intermediate options inside nested choices.
- `DeepPathTest` — deep feature call (`->>`) through an alias of a choice resolves.
- `FunctionGeneratorAliasTest` — feature call on an alias-of-choice input generates and compiles.
- `ChoiceValidatorTest` — overlap is detected through an alias of a nested choice.


.gitignore
```diff
@@ -12,6 +12,7 @@ Thumbs.db
 plugin.xml_gen
 **/*.iml
 .classpath
+.project
```
davidalk, line 15: I'm removing all the eclipse `.project` files from the repo and leaving them untracked in `.gitignore`. The reason for that is I'm using the `jdtls-lsp` plugin in Claude to allow it to navigate and manipulate Java files more efficiently. This has the effect of doing the equivalent of running `Eclipse -> Maven -> Update Project` and was causing diffs in the project files as we haven't been using eclipse in a while. It was updating `rune-integration-tests/.project` and adding `rune-generator-api/.project`. I don't see any point keeping these as we're migrating and if we want them to use the repo in Eclipse ever you can just manually run update in Eclipse and they will generate/update.

rune-lang/src/main/java/com/regnosys/rosetta/types/RChoiceType.java
```diff
@@ -114,9 +114,12 @@ public List<RChoiceOption> getAllOptions() {
...
 			return getOwnOptions().stream().flatMap(o -> {
-				if (o.getType().getRType() instanceof RChoiceType) {
-					RChoiceType nested = (RChoiceType) o.getType().getRType();
-					return Streams.concat(Stream.of(o), nested.doGetAllOptions(visited));
+				RType stripped = o.getType().getRType();
+				while (stripped instanceof RAliasType) {
+					stripped = ((RAliasType) stripped).getRefersTo();
+				}
+				if (stripped instanceof RChoiceType) {
+					return Streams.concat(Stream.of(o), ((RChoiceType) stripped).doGetAllOptions(visited));
```
SimonCockx, line 121: Reuse typeSystem.stripFrom...

rune-lang/src/main/java/com/regnosys/rosetta/types/SubtypeRelation.java
```diff
@@ -78,16 +78,16 @@ public boolean isSubtypeOf(RType t1, RType t2, boolean treatChoiceTypesAsDataTyp
 			RType t1_ = t1;
 			RType t2_ = t2;
 			return ((RChoiceType)t2).getOwnOptions().stream().anyMatch(t -> safeIsSubtypeOf(t1_, t.getType().getRType(), false, t2_, visited));
+		} else if (t1 instanceof RAliasType) {
+			return safeIsSubtypeOf(((RAliasType)t1).getRefersTo(), t2, treatChoiceTypesAsDataTypes, t1, visited);
+		} else if (t2 instanceof RAliasType) {
+			return safeIsSubtypeOf(t1, ((RAliasType)t2).getRefersTo(), treatChoiceTypesAsDataTypes, t2, visited);
```
SimonCockx, line 84: Is this necessary? If not, can be reverted.

rune-lang/src/main/java/com/regnosys/rosetta/validation/ChoiceValidator.java
```diff
@@ -55,8 +57,12 @@ public void checkChoiceOptionsDoNotOverlap(Choice choice) {
 		RChoiceType t = rObjectFactory.buildRChoiceType(choice);
 		Map<RMetaAnnotatedType, RChoiceOption> includedOptions = new HashMap<>();
 		for (RChoiceOption opt: t.getOwnOptions()) {
-			if (opt.getType().getRType() instanceof RChoiceType) {
-				((RChoiceType) opt.getType().getRType()).getAllOptions().forEach(o -> includedOptions.put(o.getType(), opt));
+			RType optType = opt.getType().getRType();
+			while (optType instanceof RAliasType) {
+				optType = ((RAliasType) optType).getRefersTo();
+			}
```
SimonCockx: Duplicate code: call strip...

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/ExpressionGenerator.xtend
```diff
@@ -395,6 +396,9 @@ class ExpressionGenerator extends RosettaExpressionSwitch<JavaStatementBuilder,
 		val receiverRType = receiverType.RType
```
SimonCockx: = `stripFromTypeAliases(receiverType.RType)`

SimonCockx: then get rid of the else branch

rune-lang/src/main/java/com/regnosys/rosetta/types/RChoiceType.java
davidalk: Agreed, will use `TypeSystem.stripFromTypeAliases`

rune-lang/src/main/java/com/regnosys/rosetta/validation/ChoiceValidator.java
davidalk: Agreed, will use TypeSystem.stripFromTypeAliases

.gitignore
davidalk: Tested this by removing all `.project` files and running `mvn clean install` so this change is fine and can stay.

rune-lang/src/main/java/com/regnosys/rosetta/types/SubtypeRelation.java
davidalk: Yes, this ordering change is necessary.

In the old ordering (`RDataType` check before `RAliasType`), consider `isSubtypeOf(Foo, FooAlias)` where `FooAlias` is an alias for `Foo`:

1. `t1.equals(t2)` → false (different RType objects)
2. Not NOTHING/ANY, not number, not string, not RChoiceType
3. `t1 instanceof RDataType` → **true** → `getSuperType()` returns `null` → returns **`false`** immediately

The `t2 instanceof RAliasType` branch is never reached, so the alias is never stripped from `t2`. The result is incorrect: `Foo` is not considered a subtype of its own alias `FooAlias`.

The same problem affects any case where `t1` is a data type and `t2` is an alias for a type in `t1`'s ancestor chain — the supertype traversal for `t1` exhausts (hits `getSuperType() == null`) before `t2`'s alias is ever stripped.

Moving the `RAliasType` checks before the `RDataType` check ensures that whenever `t2` is an alias, its underlying type is resolved first before continuing the data-type supertype traversal.

rune-lang/src/main/java/com/regnosys/rosetta/types/RChoiceType.java
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/validation/ChoiceValidator.java
davidalk: Done

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/expression/ExpressionGenerator.xtend
davidalk: Done

SimonCockx (approved):

PR #1306, opened 2026-07-03
Read/write RuneConfigurationService, config-model cleanup, ingest highlighting, read-only-namespace docs

SimonCockx:
Follow-up gaps identified while integrating the schema-import (model-import) feature.

### Read/write `RuneConfigurationService` + config-model cleanup
- New `RuneConfigurationService` reads and writes `rune-config.yml`. A roundtrip is **lossless for the properties the model knows about** and produces clean YAML — **no `null`, no empty `{}`/`[]`**. Deprecated/unknown keys (e.g. `generators.tabulators`) are dropped on read. The `ObjectMapper` is internal; callers only use `read`/`write`.
- Inlined the Jackson binding annotations onto the config POJOs and **deleted the five `*Mixin` classes**.
- `generators.namespaces` is now a plain `List<String>` instead of a `Predicate`; the namespace-matching predicate is **derived** (and no longer serialized), and `NamespaceFilter` is folded into `RuneGeneratorsConfiguration`.
- The derived `getReadOnlyNamespaces()` / `getNamespaceFilter()` are `@JsonIgnore`, so they never leak into the written file.
- Added a `RuneConfiguration` builder / `toBuilder()` so a `namespaceConfig` entry can be added (upsert by id) without rebuilding the whole config; `RuneConfiguration` stays immutable.
- `FileBasedRuneConfigurationProvider` now reads through the service.

### Highlight `ingest`/`enrich`/`projection` as annotations
They appear only inside a `[...]` transform annotation, so they now use the `variable.annotation` scope like the other annotation bodies instead of a standalone keyword.

### Read-only-namespaces docs
Updated for the `namespaceConfig` model (entries with `readOnly: true`) instead of the old top-level `readOnlyNamespaces` list.

### Tests
New `RuneConfigurationServiceTest` covers the lossless/clean roundtrip, the builder (add + upsert), and the derived predicate. Existing config tests pass.


JayasriR (approved):

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
```diff
@@ -0,0 +1,78 @@
...
+
+	public RuneConfiguration read(InputStream input) throws IOException {
+		return mapper.readValue(input, RuneConfiguration.class);
+	}
+
+	public RuneConfiguration readString(String yaml) throws IOException {
+		return mapper.readValue(yaml, RuneConfiguration.class);
+	}
+
+	public void write(Path path, RuneConfiguration configuration) throws IOException {
```
JayasriR, line 67: One small thing is it is possible to give a wrong path here, not sure how we can prevent this in dsl layer though

PR #1324, opened 2026-07-14
fix: Windows compatibility — formatter line endings, URL handling, LF checkout

SimonCockx:
## Summary

Part of a cross-repo Windows-compatibility effort (rune-common, rune-testing, rosetta-code-generators, rosetta-components, rosetta-products — see REGnosys/rosetta-products#6854).

- **`ImportManagementService`**: build the sorted import block with `\n` instead of `System.lineSeparator()`. The text is inserted into `.rosetta` documents (stored LF); on Windows the formatter/quick-fix produced CRLF import blocks inside otherwise-LF documents, bypassing the formatter's configured separator.
- **`InMemoryJavacCompiler`**: convert classpath URLs via `toURI()` instead of `URL.getFile()`, which keeps URL escaping (a space becomes `%20`) — compiled-class lookup failed for local Windows checkouts under paths containing spaces (e.g. `C:\Users\First Last\...`). Doesn't show on the Windows CI runner because its workspace path has no spaces.
- **`MultiConfigUnionTest`**: close the `URLClassLoader` so `@TempDir` cleanup isn't blocked by the open handle on Windows.
- **`.gitattributes` pinning LF checkout.** All tracked text files are already LF in the index, so no churn.

### Line-ending convention analysis

The repo previously handled line endings reactively: ~30 test call sites did `replace("\r\n", "\n")` before comparing generated code with fixtures. Two independent causes existed: (a) fixtures check out as CRLF on Windows (`autocrlf`), and (b) legacy Xtend rich-string generators emit the *runtime* platform separator. This PR now addresses both cleanly (review feedback): `.gitattributes` removes (a), and `CodeGeneratorTestHelper.generateCode` normalises generated code to `\n` once at the source for (b) — the scattered call-site normalisations are deleted. Normalisation remains only where output does not flow through that helper. Pinning the Xtend `StringConcatenation` delimiter itself (which would make *shipped generator output* byte-deterministic across OS, not just tests) is left as a separate discussion.

Verified with a full local `mvn clean install` + test run (1655 tests green), and the existing `windows-latest` job in check-pr.yml exercises this PR on Windows.

## Type of change

- Bug fix (non-breaking change which fixes an issue) — note: on **Windows**, formatted import blocks change from CRLF to LF; unix behavior is unchanged.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Sources and expectation files are stored with LF; generator tests compare
+# generated code against fixtures, so check everything out with LF on every
+# platform (overrides core.autocrlf on Windows). This also keeps Xtend
+# template sources LF so locally built generators behave like CI builds.
+* text=auto eol=lf
--- a/rune-lang/src/main/java/com/regnosys/rosetta/utils/ImportManagementService.java
+++ b/rune-lang/src/main/java/com/regnosys/rosetta/utils/ImportManagementService.java
@@ -17,6 +17,8 @@ import jakarta.inject.Inject;
 import org.eclipse.emf.common.util.ECollections;
 import org.eclipse.emf.common.util.EList;
 import org.eclipse.xtext.naming.IQualifiedNameProvider;
+import org.eclipse.xtext.nodemodel.INode;
+import org.eclipse.xtext.nodemodel.util.NodeModelUtils;
 import org.eclipse.xtext.naming.QualifiedName;
 
 public class ImportManagementService {
@@ -107,6 +109,7 @@ public class ImportManagementService {
 	//TODO: adding a new line between package blocks is a whitespace operation move this to XtextResourceFormatter
 	public String toString(List<Import> imports) {
 		StringBuilder sortedImportsText = new StringBuilder();
+		String lineSeparator = detectLineSeparator(imports);
 		
 		Import previousImport = null;
 		for (Import imp : imports) {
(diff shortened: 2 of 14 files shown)


rune-lang/src/main/java/com/regnosys/rosetta/utils/ImportManagementService.java
```diff
@@ -115,14 +115,14 @@ public String toString(List<Import> imports) {
...
-					sortedImportsText.append(System.lineSeparator());
+					sortedImportsText.append("\n");
 				}
 			}
 			sortedImportsText.append("import ").append(imp.getImportedNamespace());
 			if (imp.getNamespaceAlias() != null) {
 				sortedImportsText.append(" as ").append(imp.getNamespaceAlias());
 			}
-			sortedImportsText.append(System.lineSeparator());
+			sortedImportsText.append("\n");
```
SimonCockx: If I am editing files in the visual code plugin on Windows, I figure by default newlines will be of the form CLRF. Doesn't this change mean I will end up with a mix of newline styles in one .rosetta file?

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Sources and expectation files are stored with LF; generator tests compare
+# generated code against fixtures, so check everything out with LF on every
+# platform (overrides core.autocrlf on Windows). This also keeps Xtend
+# template sources LF so locally built generators behave like CI builds.
+* text=auto eol=lf
```
SimonCockx, line 5: Does this mean some of the existing line normalizations can now be deleted? If so, please simplify.

rune-lang/src/main/java/com/regnosys/rosetta/utils/ImportManagementService.java
SimonCockx: Good catch — fixed in 0d76bc547. `toString(imports)` now detects the line separator from the document being edited (via the node model) and reuses it, so a CRLF document gets a CRLF import block and an LF document gets LF. Hardcoding either `\n` or the platform separator could produce mixed endings; matching the document is the only choice that never does.

.gitattributes
SimonCockx: Yes — simplified in 0d76bc547. Generated code is now normalised to `\n` once at the source (`CodeGeneratorTestHelper.generateCode`; Xtend templates emit the platform separator at runtime, which .gitattributes cannot influence), and the ~30 ad-hoc `replace("\r\n", "\n")` calls that made redundant are deleted, as is the expectation-side normalisation in the regression framework (now guaranteed LF by .gitattributes). Normalisation remains only where output does not flow through that helper (FunctionGeneratorHelper's own fsa, TypeCoercionTest, ModelGeneratorUtilTest) plus the robustness normalisation in InMemoryJavacCompiler/RosettaGrammarUtil. Applied the same policy across the other PRs: rune-testing keeps its assertion-helper normalisation because it compares expectation files from consumer model repos (CDM/DRR), which its own .gitattributes cannot cover; rosetta-code-generators keeps the json-schema one (Xtend runtime output); rosetta-products/rosetta-components had none to delete.

SimonCockx:
The Windows CI failure is fixed. Root cause: `.gitattributes` switched the fixture checkout to LF, which exposed two tests that had only been passing because `core.autocrlf` happened to make their CRLF sides match: the formatter tests (Xtext's default line separator is the *platform* one, so it injected CRLF whitespace into LF documents) and `LabelProviderGeneratorTest` (raw fsa output vs fixture). The formatter now defaults to the document's own line separator — the same principle as the import-block fix above, applied to the whole formatter, with a CRLF regression test — and the label-provider test normalises its raw fsa side.

davidalk (approved):

PR #1337, opened 2026-07-24
Disambiguate the model rune-config via a build-generated marker

davidalk:
## Problem

A model's default serialisation format is looked up via `ClassLoader.getResource(...)` over the well-known `rune-config.yml`/`rosetta-config.yml` names, which returns the first match in classpath order. A child model that ships no config of its own silently picks up an ancestor's config and inherits its `defaultSerialisationFormat` without ever opting in.

This PR is the `rune-dsl` half of the fix: `rune-maven-plugin` now emits a per-model marker recording whether *that model* ships its own config, so consumers (`rune-testing`) can stop inferring presence from classpath order. It also resolves the config file location by convention instead of requiring every model to declare it explicitly.

A marker alone only fixes the "does this model have config" question. It doesn't fix a second, related failure mode: when *several* models are on the same test classpath (a model and its dependencies), the wrong marker can still win if a consumer pom happens to list an ancestor model jar before its own. This PR closes that gap too: each marker now also records the model's own identity and its direct model parents, so the winner can be computed from the model graph itself instead of classpath order.

## Changes

**`ModelPropertiesWriter`** — writes `META-INF/rune/model.properties` under a given output directory:

```properties
runeConfigPresentInModel=true
runeMavenPluginVersion=10.4.0
modelSourceGav=org.finos.cdm:cdm-java:6.23.0
modelId=org.finos.cdm:cdm-parent
parentModels=com.regnosys.rune-fpml:parent
```

- `runeConfigPresentInModel` is a filesystem check, not a classpath lookup, so it can never pick up an ancestor's file.
- `runeMavenPluginVersion` is `MojoExecution.getVersion()`, letting a consumer verify plugin/marker version relationships across the dependency graph.
- `modelSourceGav` is the generation module's own GAV — diagnostics only, so log/error text can name a real artifact instead of a URL.
- `modelId` is the model's repo identity: `getProject().getParent()`'s `groupId:artifactId` (falling back to the module's own GA if there is no parent pom).
- `parentModels` is the comma-separated GAs of the model's *direct* model parents, empty for a root model (e.g. rune-fpml). `modelSourceGav`/`modelId`/`parentModels` are all omitted (not written as blank) when ancestry computation fails, producing a marker that consumers recognise as pre-ancestry and fall back on.

**`ModelAncestry`** (new) — a pure, Mojo-free static class computing everything the marker's new keys need:

- `computeModelId` — the module's Maven `<parent>` GA.
- `parseDirectParents` — reads the model's declared parents off the `rosetta.parent.<name>.{groupId,artifactId,version}` properties every model repo already carries in its top-level pom (flattened into the module's effective properties by pom inheritance, so no extra resolution is needed); groups missing a coordinate are warned about and skipped.
- `readJarMarker` — reads a classpath jar's `modelId`/`parentModels` back out of its marker, returning empty for a jar with no marker or one missing `modelId` (pre-ancestry).
- `crossCheckClasspathModels` — warns (never fails the build) when a model jar on the compile classpath isn't accounted for by this model's declared parents or by another classpath model's own marker — a sign the hand-maintained `rosetta.parent.*` properties have rotted.

**`AbstractRuneGeneratorMojo`**:

- New `findConventionalConfigFile(File basedir)` — the single definition of "where a model's config lives": `src/main/resources/rune-config.yml`, falling back to the legacy `rosetta-config.yml`, both probed on disk. Used by both the marker's presence check and config resolution, so they can never diverge.
- `resolveConfig()` reworked: if the deprecated `runeConfig`/`rosettaConfig` parameters are set, they're still honoured (`runeConfig` wins if both are set) but now log one shared deprecation warning telling users to remove the parameter and rely on convention. If neither is set, the conventional file is resolved automatically. A conventional file found under the legacy `rosetta-config.yml` name gets its own separate warning nudging a rename.
- Both `runeConfig` and `rosettaConfig` parameters are now `@Deprecated` with matching javadoc pointing at the convention-based replacement.
- New `protected boolean writesModelProperties()` hook (default `false`) controls whether a Mojo emits the marker; `RuneGenerateMojo` overrides it to `true` (`RuneTestGenerateMojo` stays at the default, since it writes to the test-classes output directory and a second marker there would compete on the classpath with the main one).
- The marker is written in `internalExecute()` after the `errorDetected`/`failOnValidationError` check, so a failed build never emits one. Writing it runs `ModelAncestry.computeModelId`/`parseDirectParents`/`crossCheckClasspathModels` over the resolved compile-classpath artifacts; any failure there degrades to a marker with no identity/ancestry keys rather than failing the build.
- New `modelPropertiesOutputDirectory` parameter (defaults to `${project.build.outputDirectory}`) so tests can redirect the write target.

## Tests

- `ModelPropertiesWriterTest` — marker content for present/absent config, correct relative path, idempotency across repeated writes, presence/omission of the identity/ancestry keys, nothing written outside the given directory.
- `AbstractRuneGeneratorMojoTest` — `findConventionalConfigFile` for the present/legacy-fallback/absent cases; `resolveConfig` for conventional resolution, legacy fallback, absent case, and the transitional behaviour where an explicit parameter is still honoured; the legacy-name warning.
- `ModelAncestryTest` — direct-parent parsing (grouping, interpolated values, no properties, incomplete group warns+skips); `modelId` from the Maven parent GA and the no-parent fallback; `readJarMarker` including the pre-ancestry-marker and markerless-jar cases; the cross-check's undeclared/declared/transitively-accounted-for/parentless-root/markerless-jar cases, always warning-only.

`mvn -pl rune-maven-plugin -am install`: 39 tests, 0 failures, checkstyle clean.

## Out of scope

The consumer side (`rune-testing` electing the winning marker instead of inferring one from classpath order) ships as a separate, paired change in that repo — `rune-testing` hard-fails without a marker at all, and needs the identity/ancestry keys to do leaf election, so the two must release together. Downstream validation (CDM, a child model, tokenovate builds) is a post-release checklist, not part of this PR.


rune-maven-plugin/src/main/java/com/regnosys/rosetta/maven/ModelAncestry.java
```diff
@@ -0,0 +1,235 @@
...
+     * <p>
+     * Markerless model jars are skipped silently: their repo identity is not knowable from the jar,
+     * and by the version-sync invariant only root models legitimately lack markers.
+     * Presence-driven only: a parentless model with no model jars on its classpath stays silent —
+     * there is no "you declared no parents" warning. Never fails the build.
+     */
+    public static void crossCheckClasspathModels(Collection<ClasspathJar> classpathJars,
+            Collection<String> declaredParents, Consumer<String> warningSink) {
+        Map<ClasspathJar, JarMarker> markers = classpathJars.stream()
+                .filter(jar -> isModelJar(jar.file()))
```
SimonCockx: `isModelJar` here does not change the outcome: the next step, `readJarMarker`, already returns empty for any jar without a marker declaring a non-blank `modelId`, so a jar with `.rosetta` entries but no marker is dropped either way.

What it does cost is a second `ZipFile` open per artifact plus a full central-directory walk (`zip.stream().anyMatch(... .rosetta)`) — `cdm-java` has tens of thousands of entries, and this runs over every resolved artifact on every `generate`. Dropping the filter and going straight to `readJarMarker` produces the identical map with one `getEntry` lookup per jar.

`isModelJar` seems worth keeping as public API regardless — it just is not earning its place on this path.

rune-maven-plugin/src/main/java/com/regnosys/rosetta/maven/AbstractRuneGeneratorMojo.java
```diff
@@ -117,29 +127,135 @@ public abstract class AbstractRuneGeneratorMojo extends AbstractXtextGeneratorMo
...
+        try {
+            new ModelPropertiesWriter().write(new File(modelPropertiesOutputDirectory), conventionalConfigFile != null,
+                    mojoExecution.getVersion(), modelSourceGav, modelId, parentModels);
+        } catch (IOException e) {
+            throw new MojoExecutionException("Failed to write model properties.", e);
+        }
+    }
+
+    private List<ModelAncestry.ClasspathJar> classpathJars() {
+        return getProject().getArtifacts().stream()
```
SimonCockx: `getProject().getArtifacts()` returns every resolved artifact, not just the compile classpath — test- and provided-scope jars are included too.

The javadoc on `ClasspathJar` ("on the model's compile classpath") and `crossCheckClasspathModels`, and the warning text itself ("is on the compile classpath but is not declared as a model parent"), all say compile. So a test-scope model jar would trigger a warning telling the model owner to declare a parent that is not actually a compile dependency.

Worth either filtering by scope here, or relaxing the wording to "resolved classpath" so the message matches what is checked.

rune-maven-plugin/src/main/java/com/regnosys/rosetta/maven/ModelAncestry.java
SimonCockx: Actually, don't keep it in the public API if it is not gonna be used.

davidalk: Done

rune-maven-plugin/src/main/java/com/regnosys/rosetta/maven/AbstractRuneGeneratorMojo.java
davidalk: Fixed in 224f698b: `classpathJars()` is now scoped to compile/provided/system via `ScopeArtifactFilter` (the non-deprecated filter `MavenProject.getCompileArtifacts()` itself delegates to), so a test-scope model jar can no longer trigger this warning. Extracted the filtering into a static `classpathJars(Collection<Artifact>)` with unit tests covering compile/provided/system inclusion and test/runtime exclusion.

SimonCockx (approved):

PR #1354, opened 2026-08-04
Apply the setup's Rune config wiring after the injector is built

SimonCockx:
## Problem

A model built through a custom Xtext setup silently loses its Rune configuration — both its **own** `rune-config.yml` and its **dependencies'** configs.

`RosettaStandaloneSetup` accepts a config file (`setConfigFile`) and a classloader over the project's compile classpath (`setClasspathClassLoader`), and `rune-maven-plugin` sets both. The classloader matters because in a Maven build the thread context classloader is the *plugin realm*, which cannot see the project's model dependencies at all.

Both were installed as an extra Guice module **inside `createInjector()`**. But virtually every subclass of the setup overrides `createInjector()` to mix in its own runtime module — `RosettaIdeSetup` and `RosettaServerSetup` here, `CDMRosettaSetup` (rosetta-code-generators) and `BspIdeSetup` (rosetta-components) downstream — and so silently dropped that binding:

```java
public final class CDMRosettaSetup extends RosettaStandaloneSetup {
    @Override
    public Injector createInjector() {
        return Guice.createInjector(new CDMRuntimeModule());   // base-class config wiring lost
    }
}
```

The provider then fell back to the plugin realm and found no config at all (`No configuration file was found. Falling back to the default configuration.`).

This became a build failure as soon as a parent model started declaring a schema marked `[externalConfig]`. `SchemaValidator` requires a matching `schemaConfig` entry, which lives in the *parent's* `rune-config.yml` and is only reachable through the project classloader. Concretely, bumping `rune-fpml` to 3.4.0 in [finos/common-domain-model#4988](https://github.com/finos/common-domain-model/pull/4988) broke CDM's `generate-json-schema-src` execution:

```
[INFO] --- rune:10.3.0:generate (generate-json-schema-src) @ cdm-java ---
[WARNING] No configuration file was found. Falling back to the default configuration.
[ERROR] ERROR:Schema 'fpml513Recordkeeping' is marked [externalConfig] but no external
        serialization configuration is configured for it (consolidated-desc.rosetta line 11)
[ERROR] ERROR:Schema 'fpml513Confirmation' is marked [externalConfig] but no external
        serialization configuration is configured for it (consolidated-desc.rosetta line 14)
```

It looked environment-specific because that execution lives in a Maven profile only CI activates — the default execution uses `RosettaStandaloneSetup` directly and worked. It reproduces anywhere with `mvn -pl rosetta-source -Pjson-schema generate-sources`.

## Changes

**`RuneConfigurationFileProvider`** — bound as a `@Singleton`, with `fileName`/`loadFromClasspath` made mutable and a `setConfigFile(String)` setter added next to the existing `setClassLoader`, so the provider can still be pointed at a config file after the injector exists.

**`RosettaStandaloneSetup`** — `createInjector()` goes back to plain `Guice.createInjector(new RosettaRuntimeModule())`; the configuration is applied from `createInjectorAndDoEMFRegistration()` instead, which no subclass overrides. The wiring therefore survives whatever a subclass does with `createInjector()`, and no downstream change is needed. The configuration is read lazily through `RuneConfigurationHolder`, so applying it after injector creation is still in time. A guard fails loudly if `RuneConfigurationFileProvider` is rebound out of singleton scope, where post-creation configuration could not take effect.

**`StandaloneSetupConfigWiringTest`** (new) — a setup subclass that replaces `createInjector()` must still read the explicit config file and union in a dependency config reachable only through the given classloader. Fails on `main` (it picks up an unrelated config from the test classpath — the same silent-wrong-config failure mode), passes here.

## Verification

- New regression test passes; reverting only `RosettaStandaloneSetup` makes it fail.
- End-to-end: this branch installed as a snapshot, then the exact failing CDM build re-run against the **released** `default-cdm-generators` 12.6.1 → `BUILD SUCCESS`, no `[externalConfig]` errors, no fallback warning, `consolidated-desc.rosetta` validated and generated.
- Full `mvn clean install`: green, 1492 integration tests + all other modules, 0 failures.

### DIFF
--- a/rune-integration-tests/src/test/java/com/regnosys/rosetta/config/StandaloneSetupConfigWiringTest.java
+++ b/rune-integration-tests/src/test/java/com/regnosys/rosetta/config/StandaloneSetupConfigWiringTest.java
@@ -1,0 +1,85 @@
+package com.regnosys.rosetta.config;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.eclipse.xtext.testing.GlobalRegistries;
+import org.eclipse.xtext.testing.GlobalRegistries.GlobalStateMemento;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.regnosys.rosetta.RosettaRuntimeModule;
+import com.regnosys.rosetta.RosettaStandaloneSetup;
+import com.regnosys.rosetta.utils.RuneConfigurationHolder;
+
+/**
+ * A setup's configured config file and classpath classloader must reach the configuration provider
+ * even when a subclass replaces {@link RosettaStandaloneSetup#createInjector()} to mix in its own
+ * runtime module — which is what the IDE/language-server setups here and every custom generator
+ * setup downstream do. When it did not, a model built through such a setup silently fell back to the
+ * default configuration and looked its dependencies' configs up through the thread context
+ * classloader, which in a Maven build is the plugin realm and cannot see the model's dependencies.
+ */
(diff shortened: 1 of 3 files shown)


SimonCockx:
Superseded by #1355, raised from the REGnosys fork.

PR #1358, opened 2026-08-05
Generate a LabelProvider per labelled type

davidalk:
## What

Generate a `LabelProvider` for every type that carries `[label ...]` annotations directly on its own, inherited, or overridden attributes — not only for transform functions and reports, whose providers are rooted at the function's *output* type and so never contain a CSV *input* type's labels.

## Why

CSV/labelled-serialisation ingest needs to bind a column to an attribute through its label. Today a `LabelProvider` is only generated per transform function or report, rooted at the function's output type. For a projection that output *is* the labelled type, so its labels resolve — but for an ingest the output is the target model type, so the input (CSV) type's labels are absent from any generated provider. This closes that gap on the generator side.

Generating per-type is also correct on its own merits: the original reason providers were restricted to functions — a report's rule source could contribute labels — disappeared when the legacy rule `as` syntax was removed, so label content is now a pure function of the type graph.

## Changes

- **Dead code removal**: deleted the rule-source map (`attributeToRuleMap`) in `LabelProviderGenerator`, which was threaded through `buildLabelGraph` but never read. Dropped the now-orphaned `BodyCorpusLabelProvider.java` expectation fixture along with it.
- **Per-type generation**: widened `LabelProviderGenerator` to emit a provider for every type (including labelled choices) with at least one direct `[label ...]`, named `<ns>.labels.types.<Type>LabelProvider` — a separate package from the existing `<ns>.labels.<Func>LabelProvider` naming, since types and functions are distinct global-uniqueness clusters and could otherwise collide. Function- and report-rooted providers are untouched and stay byte-identical, continuing to exist alongside their type-rooted counterparts (including the case where a function's output type is itself labelled, which produces two providers with identical content by construction).
- **Discovery**: `ModelObjectGenerator` now stamps `@RuneLabelProvider` on a labelled type's generated pojo interface, computed via the same gate `LabelProviderGenerator` uses, so the annotation and the generated class can never disagree.
- **The two providers are independent, not a supersession**: function/report-rooted providers are *not* deprecated. They serve transform functions and reports permanently, and coexist with type-rooted providers, which serve a different consumer — serialisers working outside a transform function, e.g. CSV read/write against an already-flat type. Neither is framed as preferred over the other; a function-rooted provider can't generally be replaced by a type-rooted one anyway, since its output type may be defined in an upstream artifact that has no way to know a downstream model roots a transform at it.
- **Non-composability is by design and documented**: a per-type provider means "labels as seen with this type as the root." A deep-path label declared on an outer type deliberately overrides what the nested type's own labels would give, so an outer type's provider and an inner type's provider disagree on purpose — pinned by tests and recorded in the generator's class javadoc so it isn't mistaken for a bug later.
- **Docs**: `rune-and-java.mdx` documents the per-type `@RuneLabelProvider` case and describes the two provider kinds as independent and permanently coexisting; `reporting.mdx` notes that labels aren't report-specific and cross-links to it.
- **Review follow-ups**: moved `LabelProviderGenerator`/`LabelProviderGeneratorUtil` into a new `generator.java.labels` package (the gate is now shared with `ModelObjectGenerator`, so the old "function" package no longer fit), switched a javadoc reference to plain text to avoid an unused import in generated function/report sources, added an end-to-end test that generates/compiles/loads a flat labelled type with no transform or report anywhere in the model and asserts the pojo's `@RuneLabelProvider` value matches the class actually generated, and corrected a comment that wrongly claimed choice options can't carry labels.

## Out of scope

Consuming these providers on the read side (matching CSV columns to attributes by label rather than name) is a separate, downstream task — this PR only makes the type-rooted providers available for that consumer to find via `@RuneLabelProvider`.


SimonCockx (commented):
## Verdict

The core change is sound and unusually well-tested. My one substantive objection is the `@Deprecated`, which I think should come out. Everything else is small.

## What I verified

| Claim | How | Result |
|---|---|---|
| `attributeToRuleMap` was genuinely dead | read pre-PR `LabelProviderGenerator.xtend` | Only constructed, passed down, never read — and `ruleService.traverse(..., [map,context\|map])` returns the `newHashMap` unmodified, so it was always empty too. Correctly deleted. |
| The deleted `report-with-rule-references` fixture was orphaned | `git grep` on the merge-base | No test on `main` referenced it. Safe. |
| Package move breaks no downstream | grep over rune-common / rosetta-products / rosetta-components / rosetta-code-generators / rune-testing, plus a GitHub-wide code search | Zero consumers of `LabelProviderGenerator`/`…Util` outside rune-dsl (only a research repo mirroring the DSL source). |
| `@RuneLabelProvider` on pojos doesn't disturb existing readers | read `rune-common` `LabelProviderResolver`, and products' `LabelBasedSampleRowCreator:143` | Both read the annotation off a `RosettaFunction`/transform class only. No behaviour change. |
| Namespace filter preserved | read `RosettaGenerator.doGenerate` | `shouldGenerate(model)` gates before any generator runs, so type-rooted providers respect `generators.namespaces` exactly like pojos. Parented models won't regenerate their parents' providers. |
| No duplicate FQNs across models | `toLabelProviderJavaClass(RDataType)` keys off the declaring type's own namespace | One declaring model per type. Clean. |
| Real generated-output growth | scanned CDM, DRR, rune-fpml, ISO-20022 | Labels exist **only** in DRR: 2119 annotations over 36 directly-labelled types; the transitive closure over `extends` adds 2 more → **38 new provider classes in DRR, zero in CDM/FpML/ISO-20022**. Existing DRR providers are 12–200 lines, so growth is negligible. |
| Cost of running the gate over every type | `RObjectFactory.buildRDataType` uses `typeCache`; `RDataType.allAttributes` is memoised | The extra work is one `anyMatch` over already-computed attributes per type. Non-issue. |
| Old 4-arg `classBody` overload still needed | grep | `MetaFieldGenerator:66,77` uses it. Correctly retained. |
| Tests pass | CI on head SHA `a343fe97` | Green on Ubuntu **and** Windows. I did not re-run locally. |

## Findings

**1. `@Deprecated` on a class the javadoc says can never be removed — I'd drop it.** (`LabelProviderGenerator.xtend:149`)

Invariant 2 in the class javadoc argues, convincingly, that function/report providers must exist permanently. Invariant 3 then says the `@Deprecated` "is a steer, not a removal notice". That is not what `@Deprecated` means to a compiler or an IDE, and the cost is real:

```
C.java:1: warning: [deprecation] A in unnamed package has been deprecated
@Ann(A.class)
```

(verified with a standalone javac repro: a class-literal reference from an annotation does warn). The generated function class references its provider in exactly that way — `FunctionGenerator.xtend:131`, `ReportGenerator.xtend:41` — so **every model with a transform or report now emits `Note: … uses or overrides a deprecated API` when compiling its own generated code**, permanently, with no removal ever planned. I checked the CDM, DRR, rosetta-products and regnosys-parent poms: no `-Werror` anywhere, so this is noise rather than breakage — but it's permanent noise plus a strike-through in every IDE, and it actively invites the deletion that three paragraphs of javadoc exist to prevent.

The convention elsewhere in this repo is `@Deprecated(forRemoval = …, since = …)` plus a milestone issue, i.e. deprecation with an expiry. This one has no expiry by construction. A javadoc note / `@see` pointing at the type-rooted provider delivers the identical steer with none of the cost.

If it stays, I'd ask for two things: split it into its own PR, and put it in the PR title — release notes for patch/minor releases are just the auto-generated list of PR titles, and "Generate a LabelProvider per labelled type" tells a downstream reader nothing about their generated code becoming deprecated.

**2. New permanent generated surface lands ahead of its consumer.** Explicitly out of scope per the description, and there's a genuine ordering constraint (the consumer lives in another repo and needs a DSL release first), so this isn't an objection — but please state the merge order in the PR, and say whether the CSV story needs this backported to 9.x. One mitigating point worth recording: because discovery goes through `@RuneLabelProvider`, the `.labels.types.<Type>LabelProvider` naming isn't frozen the moment it ships — only hand-written references would pin it. That argues for landing the consumer soon rather than for blocking here.

**3. `createTypeRepresentation` casts silently.** `LabelProviderGenerator.xtend:107-112` and `:127` do `target as RDataType` in the `else` branch. Today `streamObjects` only yields `RFunction` and `RDataType`, so it's correct; a third `RObject` kind added later gets a `ClassCastException` from inside a generator rather than a clear failure. An explicit `else throw new IllegalStateException(...)` makes it loud instead of silent.

**4. Optional: the generator is now two generators wearing one coat.** Root selection, class naming and the header all branch on `instanceof RFunction`, and the two flavours share only `buildLabelGraph`. Splitting into a function-rooted and a type-rooted generator over a shared graph builder would remove every `instanceof` and turn the eventual fluent-API migration into two small ports instead of one branchy one. Not blocking — and the byte-exact `type-flat/FooLabelProvider.java` fixture is a good parity gate for whenever that happens.

## Things done well

- `TypeLabelProviderDiscoveryTest` is exactly the right shape: generate → compile → load → assert the pojo's annotation *is* the class that was actually emitted. That's the one drift neither generator's own test could catch.
- `assertGeneratedFiles(...)` asserting the exact file set (rather than "exactly one file") means a future over-broad gate fails loudly instead of silently multiplying output.
- The "direct labels only" gate with the rejected alternative recorded in the javadoc, and the deliberate outer/inner disagreement pinned by `testOuterTypesDeepPathLabelDeliberatelyDisagreesWithInnerTypesOwnLabel`, are the right way to stop a later "simplification".
- Function/report provider output stays byte-identical apart from the added header.

## Coverage gap I'd call acceptable

No model under `generation-regression-tests` carries a `[label` at all, so the regression corpus contributes zero coverage for this path — the dedicated tests are all of it. Given how thorough they are I wouldn't ask for a corpus model, but it's worth knowing that's where the safety net is. The one interaction no test covers is the namespace filter (a labelled type in an excluded namespace must not emit a provider); I verified that by reading `RosettaGenerator`, and since it's a model-level gate ahead of every generator it's structurally safe rather than something the new code could get wrong.

davidalk:
Addressed finding 1 in c807df66, keeping `@Deprecated` but giving it a real expiry instead of implying it never goes away.

The function/report providers genuinely aren't permanent — they're a fallback for the case where an upstream artifact's dependency chain hasn't yet been rebuilt with a DSL version that emits type-rooted providers. That gap closes at the next major version, once every artifact in the chain has had a full major cycle to regenerate. Before this change the javadoc said "permanently, not just until some future cleanup," which is exactly the "no expiry by construction" problem you called out — deprecated with no plan to ever remove it.

Reworded invariants 2 and 3 in the class javadoc, the generated `@deprecated` javadoc itself, the three expectation fixtures, and the doc note in `rune-and-java.mdx` to say: removal is scheduled for the next major version, not before, and here's why it can't happen sooner. Kept `@Deprecated` itself (no `forRemoval`/`since` attributes) since we don't have a concrete version number yet, but the message now reads as a real deprecation with an end date rather than a permanent steer.

SimonCockx:
Thanks — the expiry is the right shape. One problem with the reason given for it.

I built the branch and ran this case:

```
type Outer:
	nested Inner (1..1)

type Inner:
	attr string (1..1)
		[label "Inner label"]

func MyFunc:
	[codeImplementation]
	[ingest JSON]
	inputs:
		inp int (1..1)
	output:
		result Outer (1..1)
```

Generated files: `test/labels/types/InnerLabelProvider.java` and `test/labels/MyFuncLabelProvider.java` — **no `OuterLabelProvider`**. And `MyFuncLabelProvider.getLabel("nested.attr")` returns `"Inner label"`, so the function-rooted provider is non-empty and load-bearing.

That follows from the gate: `LabelProviderGeneratorUtil` is deliberately "direct labels only", `Outer` has no label of its own, so **no DSL version with the current gate ever emits a type-rooted provider for it**. Regenerating every artifact doesn't change that.

So there are two gaps with different lifetimes:

- **Rebuild lag** — a dependency built by an older DSL has no `labels.types` package at all. Closes once everything regenerates; your argument holds here.
- **Gate scope** — output type with no direct labels but labelled descendants. Structural, and no rebuild touches it.

Invariant 2 still says this two paragraphs above the new text — *"no amount of widening the gate fixes this"* — so the javadoc now argues the blocker is structural in one place and temporal in another.

That makes it riskier than the honest "permanent" version it replaced: whoever works the milestone at the next major will read "scheduled for removal", delete the class, and models whose transform output types carry only nested labels lose their labels silently.

The expiry needs a mechanism, not just a date. Either:

- widen the gate to "has a reachable label" — recorded in `LabelProviderGeneratorUtil` as considered and rejected on output-size grounds, which is worth revisiting with an actual measurement; or
- add a consumer-side fallback — resolve the type-rooted provider, fall back to the function-rooted one.

Whichever it is, that's what the milestone issue should carry and what the javadoc should point at, instead of "by then everything will have regenerated".

Minor: `forRemoval = true` needs no version number and is the machine-readable half of the commitment — worth adding even without a concrete release.

davidalk:
Thanks Simon — the `Outer`/`Inner` example makes the problem concrete: with the direct-labels-only gate, no DSL version ever emits an `OuterLabelProvider`, so "scheduled for removal at the next major version" was pointing at a gap that regeneration can't close. Good catch.

As agreed, rather than give the expiry a real mechanism (widen the gate, or add a resolver-side fallback), we're dropping the deprecation entirely, in c807df66..723347c4:

- **The function/report-rooted provider is permanently necessary** for transform functions and reports — not a stopgap awaiting a type-rooted replacement. It's unaffected by whether the output type happens to also qualify under the type-provider gate, so there's no version at which it becomes safe to remove.
- **The type-rooted provider's narrow gate stays as-is**, but for a different reason than before: it isn't a scope limitation we were planning to widen, it's the correct scope for what that provider is *for* — serialiser consumers working outside a transform function (e.g. labelled CSV read/write against a type that's already flat, which is the story this task exists to support). Widening it to "reachable label" would only be solving a problem that the function/report provider already handles for the transform-output case.

So the two providers coexist permanently, generated independently, with neither positioned as a replacement for the other. `@Deprecated`, the `@deprecated` javadoc, and invariant 3 in `LabelProviderGenerator`'s class javadoc are all removed; the three function/report expectation fixtures are back to their pre-deprecation content; `LabelProviderGeneratorUtil`'s gate javadoc now explains the serialiser-consumer rationale directly instead of leaving it implicit. `mvn -o install` and the focused `LabelProviderGeneratorTest` (10/10) are green.

SimonCockx (approved):
Re-reviewed the delta since my last pass (`a343fe97` → `723347c4`). Approving.

**Finding 1 is resolved, and resolved the right way.** The `@Deprecated` is gone rather than given a nominal expiry, and the reason recorded in `LabelProviderGenerator`'s invariant 2 is the correct one: the gate is deliberately "direct labels only", so an output type whose labels all sit on nested descendants never gets a type-rooted provider in *any* DSL version. That's structural, not staleness — no amount of regenerating dependencies closes it, so there was never a mechanism behind "removed in the next major". Better to say permanent and mean it.

Verified on the diff: function- and report-rooted provider output is byte-identical to `main` again (the three expectation fixtures lose exactly the header they gained), so the permanent `[deprecation]` note in every downstream model build goes away with it. The `rune-and-java.mdx` / `reporting.mdx` wording now describes the two providers as independent rather than one superseding the other, which matches.

**Findings 2 and 4** — merge order is now stated (finos/rune-common#702 carries it explicitly), and the generator split stays a "whenever someone migrates it" note, which is fine.

**Finding 3 is still open** and still minor: `LabelProviderGenerator.xtend:108` and `:124` cast `target as RDataType` in the `else` branch. Correct today because `streamObjects` only yields `RFunction` and `RDataType`; a third `RObject` kind added later gets a `ClassCastException` from inside a generator instead of a clear failure. An explicit `else throw new IllegalStateException(...)` would make it loud. Not blocking.

One note for the release, not for this PR: I've raised a blocking finding on the consumer side in finos/rune-common#702 — its guard drops the function-rooted provider for reports, enrichments and pre-annotation models. Nothing here needs to change for it, but the two shouldn't be bundled into a release as a matched pair until that's settled.

SimonCockx (approved):

PR #1359, opened 2026-08-05
Mark importer-generated namespaces, and report edits to them

SimonCockx:
> #1360, which split the namespace-pattern unification out of this PR on review, has merged. This branch is rebased onto it, so the diff below is now this PR's own change only.

## What

A namespace produced by an importer from data *samples* is editable by design: samples carry no constraints, no complete enumerations and no documentation, so the import produces a working stub that a modeller refines. Read-only is the wrong marker for it, and a blocking check is the wrong guard — one that gets bypassed on nearly every pull request stops meaning anything.

Groundwork for importing Rune types from CSV samples: an importer will produce namespaces from data rather than from a schema, and those need a different guard from read-only ones. Reference implementation: [rosetta-models/demo#483](https://github.com/rosetta-models/demo/pull/483) — a purpose-built branch, not demo's `main`; no production model has adopted the marker yet.

## The origin marker

`RuneNamespaceConfiguration` gains `origin`, keyed on the producing tool and independent of `readOnly` and `schemaConfig`:

```yaml
namespaceConfig:
- namespace: demo.unavista.csv
  origin:
    modelImport: csv
```

`readOnly` is absent because a sample-derived namespace is editable; the marker is separate from `schemaConfig` because a hand-written type may legitimately carry a schema configuration without any import.

**The value is a closed field, not an open map.** An earlier round backed it with a
`Map<String, String>` so an unrecognised tool key would survive a read/write roundtrip. That is
unnecessary: a configuration is only ever read by the version that wrote it or a newer one — a model
reads its dependencies' configs, never the reverse — so a reader always knows every key a config can
contain. An unrecognised key is therefore a mistake, and is dropped like any other unknown property.
Adding a second tool later means adding a field, shipped in the same version as the checker that
reads it.

`origin: {}` names no tool and so says nothing about provenance — and is not written back as noise.

## The annotation

The reusable workflow gains a step — not a second job, so a consumer gets no extra checkout and no extra check name. It never fails on a model edit: one annotation per changed file inline in *Files changed*, plus a summary table.

Classification is deliberately **fail-safe**: recognising a list of risky constructs would fail open on anything it had not been taught, so it recognises only what is provably harmless.

| Change | Level |
|---|---|
| Leaves the same set of declarations — documentation, comments or reordering | notice |
| Only adds declarations — data that parsed before still parses | notice |
| Modifies or removes a declaration, **or deletes the file** | **warning** |

Three cases the first draft got wrong, all now reproduced end-to-end against throwaway repositories:

- **A deleted file was skipped entirely** — it fails `os.path.exists`, so the most consequential edit was the only silent one. Deletions are now read from the base ref and always warn.
- **A rename plus a removal read as purely additive** — diffing only the destination path makes git render a rename as a wholly new file. `classify` now passes both pathspecs, and the finding says the file was renamed.
- **`//` inside a string literal truncated the line** — changing `[label "http://old.example.com/a"]` classified as documentation-only. Comment stripping is now string-aware.

A failed git invocation used to leave the job silent, indistinguishable from "nothing changed". It now emits a `::warning::` saying the check could not run — still without failing, since this check never blocks.

## Type of change

- **New feature** (non-breaking).
- **This change requires a documentation update** — a `generated-namespaces` page, cross-linked with the read-only one, plus the stale `rune-docs.netlify.app` links corrected to `rune.finos.org` in `docusaurus.config.ts`, `website/README2.md` and the workflow comment.

## Verified

- Full `mvn clean install` green; 46 Python unit tests.
- The annotator run end-to-end against throwaway repositories: the demo PR's outcome reproduced exactly, plus the deletion, rename-and-remove, URL-label and broken-base-ref cases above.
- The bundle (rune-common, rune-testing, rosetta-code-generators, bsp, rosetta-translate, model-import) builds against this branch.
- Netlify preview green, so the new page and its cross-links resolve (`onBrokenLinks: 'throw'`).

## Test dispositions

| Production file | Covered by |
|---|---|
| `RuneOriginConfiguration` | `RuneConfigurationServiceTest.anUnrecognizedOriginKeyIsReadAsGeneratedAndSurvivesARoundtrip` and `anEmptyOriginSaysNothingAboutProvenance` — the open map read, written back and re-read |
| `RuneNamespaceConfiguration` | `originMarkerIsReadAndRoundtrips`, which also asserts `readOnly` stays absent through a write, and `aNamespaceCanBeReadOnlyWithoutCarryingAnOrigin` |
| `RuneConfiguration` | unchanged by this PR now that `getGeneratedNamespaces()` is gone |
| `annotate_generated_namespaces.py` | `test_annotate_generated_namespaces.py` — origin parsing including an unknown tool, string-aware comment stripping, every classification outcome plus deletion, rename and an unrecognised construct, root filtering, finding selection and ordering, and the git-failure path |
| `verify-readonly-namespaces.yml` | a workflow definition; not unit-testable. The script it runs is covered above and was run end-to-end |
| `generated-namespaces.mdx`, `sidebars.ts`, `docusaurus.config.ts` | documentation and a site URL; no unit test applies. All three are exercised by the Netlify build on this PR, which fails on a broken link |

## Addressed from review, deferred

- **Every accessor without a caller is gone.** `isGeneratedNamespace`, `isReadOnlyNamespace`, `getGeneratedNamespaces()` and `RuneNamespaceConfiguration.isGenerated()` were all speculative surface. What is left is only what models the file format: `getOrigin()` and `getModelImport()`. Whichever change first needs a derived helper adds it there.
- **Backport to `9.x.x`** — that line has `namespaceConfig` but not this feature, and consumers on it also track the `@main` checker. Worth a decision, but not in this PR.
- **Downstream consumers** that forward to `RuneNamespacePattern` cannot compile until a release carrying #1360 reaches them. Order: #1360 → #1359 → DSL release → downstream version bump.


SimonCockx (commented):
Reviewed at `8a357f8`. The design reasoning holds up — a sample-derived namespace genuinely is a different thing from a schema-derived one, and reporting rather than blocking is the right call for it. The refactor is clean and the compatibility story is sound.

Three things do not do what the PR says they do, and they share a shape: the mechanism silently does nothing in exactly the parts described as fail-safe.

## What I verified, and how

|---|---|---|
| The generators filter is unchanged by the refactor | Compiled the verbatim pre-PR `buildNamespaceFilter` alongside `RuneNamespacePattern`, compared over every pattern/namespace combination incl. pattern pairs and the empty list | **0 differences** |
| Java and Python matchers agree | Same case grid through both | **2 disagreements** (§3) |
| `origin` survives read and write | Ran `RuneConfigurationService` on a config carrying an unrecognised origin key | **Marker destroyed** (§1) |
| The annotator handles real diffs | Throwaway git repos: delete, rename+removal, label change behind a `//` | **Three mishandled** (§2) |
| Existing configs round-trip unchanged | Read + write a pre-change config | No `origin` key emitted — clean |
| An older DSL can read a config carrying `origin` | Checked the reader at tag `10.1.0` and on `9.x.x` | `FAIL_ON_UNKNOWN_PROPERTIES=false` on both — forward compatible |
| rune-runtime binary compatibility | 4-arg `RuneNamespaceConfiguration` ctor retained with an identical descriptor; located the one external production caller (`SchemaImportServiceImpl`) | Safe |
| `origin` survives the classpath union | `FileBasedRuneConfigurationProvider` passes entries by reference | Safe |
| Consumer blast radius | Fetched all 10 calling repos' workflow inputs and `rune-config.yml` | **Zero affected today** (§4) |
| CI layout resolves the sibling import | Reproduced the `.rune-dsl-checker/` layout, invoked from the repo root | Works |

## 1. An unrecognised `origin` key is dropped — and a round-trip erases it

The javadoc and the docs page both sell the marker as extensible ("keyed on the producing tool, so a future generator can record its own provenance"). It is backed by one hardcoded field. With `origin: {someFutureTool: v1}`:

- Java reads it as **not generated** — `getGeneratedNamespaces()` returns `[]`, no error.
- Python reports it as **generated**, tool `someFutureTool`.
- A round-trip through `RuneConfigurationService` rewrites it as **`origin: {}`**.

```
in:  origin:
       someFutureTool: v1
out: origin: {}
```

The component that *writes* the file is the one that deletes the data, and model-import writes through this service. Same across versions: a newer DSL writes an origin key this version doesn't know, this version reads and rewrites the file, the key is gone — and `origin: {}` then reads as not-generated on both sides, so the annotation quietly stops firing.

Either back it with a `Map<String, String>` (or `@JsonAnySetter`/`@JsonAnyGetter`), or drop the extensibility framing and reject an unknown key loudly. As it stands the promise is unbacked in the one direction that loses data.

## 2. The annotator misses the three changes that most need annotating

All three reproduced end-to-end against throwaway repositories.

**Deletion → no annotation at all.** `findings_for` skips any path failing `os.path.exists` (`annotate_generated_namespaces.py:130`). Deleting a whole generated type is the most consequential edit and the only one that is silent.

**Rename + removal → a `notice` saying nothing was removed.** I renamed a file, dropped `amount number (1..1)` and renamed `tradeIdentifier` → `newIdentifier`. `classify` diffs a single destination pathspec, so git cannot detect the rename and renders it as wholly new:

```
::notice … This pull request has added declarations, without changing or removing any.
```

A renamed label is precisely what the warning text exists to catch, and this is worse than silence — it is an active all-clear.

**A `//` inside a string literal defeats the classifier.** `LINE_COMMENT_RE` strips from `//` to end of line regardless of string context. Changing `[label "http://old.example.com/a"]` to `"http://new.example.com/b"` yields:

```
::notice … This pull request has changed documentation or comments only.
```

The fix for the first two exists next door: `check_readonly_namespaces.changed_rosetta_files` already yields `(status, old, new)` from `git diff --name-status -z` and reads the base version via `git_show`. That helper was not moved into the shared module — it should be, and `classify` should diff both pathspecs.

## 3. The two matchers had already diverged before merge

Bare `*`: Java matches everything (`parent()` of a single segment is empty, and `startsWith(empty)` is always true); Python matches nothing (`"*"` doesn't end in `".*"`, so it falls through to equality). Under the read-only checker that also trips the stale-pattern guard, so the build fails with *"stale or mistyped pattern?"* while the actual violation goes unreported.

The PR's central claim is one interpretation, and the mechanism holding them together is a comment saying they must be kept in step — which has already failed at introduction. The durable fix is a shared corpus of cases both sides read, so a divergence fails a test rather than waiting to be noticed.

## 4. The read-only checker's behaviour change is understated, and ships on merge

The body lists two disagreements with `fnmatch`. There are more: bare `*`, `foo*` (matched `foo`, `foobar`, `foo.bar` → now nothing), `a?c`, `a[bc]`. All glob syntax other than a terminal `.*` silently stops working.

Blast radius is **zero today** — of the 10 repos calling the reusable workflow, 8 have no `namespaceConfig` at all; ISO-20022 has 16 read-only patterns and rune-fpml 2, every one an exact namespace with no wildcard. But it reaches all 10 the moment this merges (`@main`, and no consumer sets `checker-ref`), so it belongs in the body as a stated behaviour change.

Worth a follow-up issue rather than a change here: rune-fpml's `fpml.consolidated` protects only 2 of that repo's 42 `.rosetta` files, the other 40 being in sub-namespaces. The fix, `fpml.consolidated.*`, is *only* correct because of this PR — under the old glob reading it would have silently dropped the 2 files that sit at the parent namespace.

## 5. A broken base ref makes the annotate job silently do nothing

`git()` uses `check=False` and ignores the return code. With `--base origin/does-not-exist` the script exits 0 with no output. Since the job never fails, nothing distinguishes "no generated namespace changed" from "the check did not run". The script already emits a `::notice::` for a missing config; a failed git invocation deserves the same.

## 6. Three new public predicates ahead of any caller

`getGeneratedNamespaces`, `isGeneratedNamespace` and `isReadOnlyNamespace` have no production caller anywhere (searched finos, REGnosys and rosetta-models — tests only). The consumer of the marker is the Python checker, which parses the YAML directly, and REGnosys/rosetta-products#7002 forwards to `RuneNamespacePattern` rather than to `isReadOnlyNamespace`. `RuneOriginConfiguration` and `RuneNamespacePattern` earn their place; these three predicates are speculative permanent surface in the module where that costs most.

## Scope

Five independently significant changes here: the `origin` marker, the matcher unification (a behaviour change to a *blocking* check on 10 repos), the annotate job, new CI for the Python scripts, and the docs-URL fix. The matcher change in particular would be worth landing alone, so it can be reverted independently of the feature.

## Minor

- The docs example renames the caller's job id `readonly-namespaces:` → `namespaces:`. A consumer following it changes their check name and breaks any branch-protection rule pinned to it. (The reusable workflow's own `verify` job id is unchanged — that's the part that matters.)
- `annotate-generated` duplicates the two checkouts, the Python setup and the base-ref resolution verbatim, adding a second `fetch-depth: 0` clone to every consumer PR — for a script that never fails, in 9 repos that have no `origin` marker. A second step in `verify` with `if: always()` would do.
- `write_summary` emits literal `warning`/`note` text in the icon column.
- Reordering declarations classifies as "changed documentation or comments only" — harmless outcome, misleading wording.
- `getGeneratedNamespaces()` returns duplicates for rune-fpml's two-entries-one-namespace shape (consistent with the existing `getReadOnlyNamespaces`, so not a regression).
- The PR body omits the template's *Type of change*, which is the signal for the version bump.
- rosetta-products#7002 cannot compile until a DSL release carrying `RuneNamespacePattern` reaches the bundle — worth stating the order explicitly.
- Does this want backporting to `9.x.x`? That line has `namespaceConfig` but not this feature, and the `@main` checker serves those repos too.
- "Running for real" is a purpose-built base branch (`demo/csv-import-annotation-base`), not demo's `main` — no production model has adopted the marker yet.

## Verdict

§1 and §2 are worth fixing before merge; both are places where the mechanism silently does nothing rather than failing, in the parts the PR describes as fail-safe. §3–6 are worth addressing but none blocks. The Java-side refactor and the compatibility story I checked hardest are clean.

SimonCockx:
Thanks — the shared shape you spotted was the right read: three places where the mechanism silently did nothing in the parts described as fail-safe. All six are addressed, and the PR is split.

## §1 — `origin` is now an open map

`RuneOriginConfiguration` holds `Map<String, String>` with a delegating `@JsonCreator` and `@JsonValue`, so an unrecognised tool key reads as generated and is written back unchanged. `origin: {}` reads as not-generated on both sides and is not written out at all.

You were right that the write side is what mattered: the component that rewrites the file was the one deleting the data. Two tests pin it — `anUnrecognizedOriginKeyIsReadAsGeneratedAndSurvivesARoundtrip` reads `someFutureTool: v1`, writes, and re-reads it as generated.

Finding this also surfaced a latent bug next door: `namespace_config_entries` was flattening indentation, so nesting was lost and `origin`'s keys were indistinguishable from `schemaConfig`'s. It now dedents while preserving structure, which also stops a nested `namespace:` being mistaken for a top-level one.

## §2 — all three reproduced, fixed, and pinned

Your reproductions were exact. I rebuilt them as throwaway repositories and they now give:

| Case | Before | Now |
|---|---|---|
| Delete a generated type | *(silent)* | `warning … has deleted the file` |
| Rename + drop `amount` + rename `tradeIdentifier` | `notice … added declarations, without changing or removing any` | `warning … renamed the file and changed or removed existing declarations` |
| `[label "http://old…"]` → `"http://new…"` | `notice … documentation or comments only` | `warning … changed or removed existing declarations` |

As you suggested, `changed_rosetta_files` moved into the shared module (as `changed_files`, with `--find-renames`), and `classify` now passes both pathspecs. Comment stripping is string-aware rather than regex-based. Deletions are read from the base ref.

Reordering now says "changed documentation, comments or ordering only" — the outcome was already right, only the wording was misleading.

## §3 — the corpus exists

`.github/scripts/namespace-pattern-cases.json`, read by `RuneNamespacePatternTest` (as dynamic tests) and `test_namespace_patterns.py`. The bare-`*` divergence is in it, resolved as *matches everything* — the reading Java already had, and the more useful one under the stale-pattern guard.

## §4 — stated, with the full list

The behaviour change is now its own PR (#1360) with a **Breaking change** type, the complete list of glob syntax that stops working, and the blast radius: zero today, all ten repositories on merge. Your rune-fpml observation is in there as a follow-up — `fpml.consolidated` protecting 2 of 42 files, and the fix being correct *only* because of this change.

## §5 — a failed git invocation is now visible

`git()` raises `GitError`; the annotator catches it and emits `::warning::The generated-namespace check could not run, so no edit was reported: …`, still exiting 0. Silence no longer means two different things.

## §6 — two of the three predicates dropped

`isGeneratedNamespace` and `isReadOnlyNamespace` are gone. `getGeneratedNamespaces()` stays: it mirrors the existing `getReadOnlyNamespaces()` and is what C2.1 reads to derive `isImported`.

## Scope — split

#1360 is the matcher change alone, revertible on its own. This PR is the marker and the annotation. I left the docs-URL fix and the Python CI with the halves they belong to rather than splitting five ways.

## Minor

Job id `readonly-namespaces:` restored in the docs example. The annotate job is now a **step** of `verify` with `if: always()`, so no second checkout and no new check name. Summary icons are ⚠️/ℹ️. Type of change added to both bodies. Release order stated on rosetta-products#7002.

Left alone: `getGeneratedNamespaces()` duplicates, consistent with `getReadOnlyNamespaces()` as you note. The `9.x.x` backport is a real question — flagged in the body, not decided here.

One thing your review did not raise that I hit: **rune-runtime targets Java 8**, so the test I first wrote with `Path.of`/`Files.readString` would not have compiled. Worth knowing for anything else landing in that module.

.github/scripts/annotate_generated_namespaces.py
```diff
@@ -0,0 +1,259 @@
...
+MAX_INLINE_ANNOTATIONS = 10
+
+# Every annotation ends with this, so a reviewer who has not seen one before can find out
+# what generated namespaces are and why the note is there.
+DEFAULT_DOCS_URL = "https://rune.finos.org/docs/developers/generated-namespaces"
+
+# `origin:` on its own line, followed by its indented tool entries. `origin: {}` does not match,
+# and so says nothing about provenance.
+ORIGIN_RE = re.compile(r'^origin\s*:[ \t]*$\n((?:[ \t]+\S.*(?:\n|$))*)', re.MULTILINE)
+ORIGIN_ENTRY_RE = re.compile(r'^[ \t]+([A-Za-z0-9_]+)\s*:\s*(.+)$', re.MULTILINE)
```
davidalk: **Java and Python read `origin` differently, in three ways — all silent**

I put nine YAML shapes through both `RuneConfigurationService` and `parse_generated_namespaces`:

| `origin` shape | Java | Python |
|---|---|---|
| `origin: {modelImport: csv}` (flow style) | generated, `{modelImport=csv}` | **not generated** |
| `origin:`<br>`  modelImport:` (no value) | generated, `{modelImport=null}` | **not generated** |
| `origin:`<br>`  someTool:`<br>`  modelImport: csv` | `{someTool=null, modelImport=csv}` | **`{someTool: 'modelImport: csv'}`** |

The third is the worst. `ORIGIN_ENTRY_RE` requires `(.+)$` after the colon, so a valueless key swallows the *following line* as its value: the annotation then reads `maintained by someTool (modelImport: csv)` and the real tool is gone. The first two make the annotator skip a namespace that Java considers generated, with no output at all.

Flow style is the realistic one — Jackson reads it happily, so any config written by hand or by a third-party tool in that style is invisible to this check.

This is §3 of the last review recurring one layer up. `namespace-pattern-cases.yml` made a `matches()` divergence fail a test; nothing plays that role for `origin` parsing, which is still two independent implementations of one file format. The same treatment applies: a corpus of `origin` fragments with their expected tool map, read by both `RuneConfigurationServiceTest` and `test_annotate_generated_namespaces.py`.

.github/scripts/annotate_generated_namespaces.py
```diff
@@ -0,0 +1,259 @@
...
+    if not generated:
+        return findings
+
+    for code, old_path, new_path in changed_files(base, "*.rosetta"):
+        path = old_path if code == "D" else new_path
+        if not under_root(path, root):
+            continue
+        # A deleted or renamed-away file is read from the base ref: it is exactly the change
+        # that most needs reporting, and it is not on disk to be read.
+        namespace = namespace_of(read(new_path) if code != "D" else None) \
```
davidalk: **A file dragged out of a generated namespace is unreported**

The `or` short-circuits, so the base ref is consulted only when the new content is *absent*. A pull request that edits the `namespace` declaration itself is therefore classified on its new namespace alone.

Reproduced end-to-end: changing `namespace demo.unavista.csv` → `namespace demo.handwritten` **and** removing `amount number (1..1)` in the same commit produces no annotation and exit 0.

That is the deletion bug one step along — the classifier is fail-safe, but the *selection* still is not. Taking the union of the base and head namespaces fixes it, and states the intent that the current code only implies: *was this file ever in a generated namespace?*

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneOriginConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+ * Provenance is deliberately separate from {@link RuneNamespaceConfiguration#isReadOnly() readOnly}
+ * and from {@link RuneNamespaceConfiguration#getSchemaConfig() schemaConfig}. A namespace generated
+ * from an authoritative schema is read-only, one inferred from samples is not, and a hand-written
+ * type may carry a schema configuration without any import having taken place.
+ */
+public class RuneOriginConfiguration {
+	/** The key model-import records its provenance under; its value is the source format. */
+	public static final String MODEL_IMPORT = "modelImport";
+
+	private final Map<String, String> tools;
```
davidalk: **`Map<String, String>` is not the open map the javadoc describes**

Two shapes, both verified by running `RuneConfigurationService` directly.

**A structured or list value fails the whole config read**, not just the marker:

```yaml
origin:
  futureTool:
    version: 2
```
→ `MismatchedInputException: Cannot deserialize value of type java.lang.String from Object value`

`FAIL_ON_UNKNOWN_PROPERTIES=false` covers unknown *keys*, not a type mismatch inside a typed map. So a future tool recording anything richer than a scalar makes `rune-config.yml` unreadable for every consumer on this version.

**A null-valued key is erased by the writer.** `origin:` / `&nbsp;&nbsp;someTool:` reads as generated (`{someTool=null}`) and then writes back with `someTool` gone — the whole `origin` dropped when it is the only key, the key dropped silently when it sits beside a real one. That is the data-loss shape from §1 surviving in a narrower form.

**Suggested fix: narrow the claim rather than widen the type.** Say in this javadoc and on the `generated-namespaces` docs page that a tool's value must be a non-null scalar, and the promise then matches what the code does. Widening to `Map<String, Object>` is the lesser option — it would make the extensibility literally true, since the marker's only semantics are "is there a key", but it costs the type safety of `getModelImport()` and buys a generality nothing is asking for yet. Either way, what should not stand is the current gap between the javadoc and the type.

.github/scripts/annotate_generated_namespaces.py
```diff
@@ -0,0 +1,259 @@
...
+import os
+import re
+import sys
+
+from namespace_patterns import GitError, changed_files, git, git_show, matches, namespace_config_entries, namespace_of
+
+DOC_STRING_RE = re.compile(r'<"(?:[^"\\]|\\.)*">')
+
+# GitHub renders at most ~10 annotations per level per step.
+MAX_INLINE_ANNOTATIONS = 10
```
davidalk: **The annotation cap drops findings silently, and the docs promise otherwise**

Verified with 14 changed files in a generated namespace: 10 inline annotations, 14 rows in the step summary, and nothing anywhere saying 4 were omitted.

The summary table is a real mitigation — a reviewer who opens the run sees everything. But `generated-namespaces.mdx:94` says *"Each changed file in a generated namespace gets one annotation"*, which stops being true at the eleventh file, and *Files changed* is where a reviewer actually looks.

Two things worth doing: say in the docs that at most 10 appear inline and the rest are in the run summary, and emit a final notice naming the count that was dropped. A cap that is invisible reads as "nothing else changed".

Minor, on the same line: the comment says GitHub renders ~10 annotations *per level*, but the slice is over the combined list — 10 warnings will suppress every notice even though the notice budget is untouched.

davidalk (commented):
The three defects from the last round are genuinely fixed — I reproduced all three end-to-end against throwaway repositories and each now behaves as described. Binary compatibility is clean: `javap -s` shows the 4-arg `RuneNamespaceConfiguration` descriptor byte-identical, and reading + writing the pre-change config emits no `origin`.

Findings inline. Separately, and not about the code:

**The PR body links to internal REGnosys resources.** `finos/rune-dsl` is public, so nobody outside the org can open either of these, and the body is permanent and indexed:

- the Notion CSV import design link (`notion.so/regnosys/…`)
- the `STORY-1936` Airtable reference

Both are worth stripping before this merges. The body already explains the design in its own terms, so replacing them with a one-line summary of what they establish loses nothing. (`rosetta-products#7002` in the last section has the same problem — it renders as a broken reference to anyone outside REGnosys.)

davidalk (changes requested):

.github/scripts/annotate_generated_namespaces.py
SimonCockx: Confirmed all three, including the line-swallowing one — `someTool:` produced `{"someTool": "modelImport: csv"}`.

Rather than pin two parsers with a corpus, I removed one: both checkers now read `rune-config.yml` with **PyYAML** instead of walking the block structure by hand. The reusable workflow already runs `setup-python` and we own it, so `pip install` is one line — and the stdlib-only constraint that caused this bought nothing, since these scripts only ever run inside our own workflow.

All three shapes now read as Java reads them, and `namespace_patterns.py` lost 47 lines (`namespace_config_entries`, `entry_value`, the indentation walker, `ORIGIN_RE`, `ORIGIN_ENTRY_RE`). Each shape has a test.

Install cost measured: 7s → 2s with a pinned `requirements.txt` and `cache: 'pip'`.

.github/scripts/annotate_generated_namespaces.py
SimonCockx: Reproduced and fixed. `findings_for` now takes the union of the base and head namespaces rather than short-circuiting on the head, which states the intent directly: *was this file ever in a generated namespace?*

The move is named in the annotation and escalated to warning:

```
::warning … has moved the file from 'demo.unavista.csv' to 'demo.handwritten'
and changed or removed existing declarations.
```

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneOriginConfiguration.java
SimonCockx: Both shapes confirmed — the nested value aborts the whole config read, and a null-valued key is erased on write.

Neither of your two options, in the end: **the open map is gone**. The premise it rested on does not hold — a configuration is only ever read by the version that wrote it or a newer one, since a model reads its dependencies' configs and never the reverse (`FileBasedRuneConfigurationProvider`). So there is no scenario where a reader meets a key it does not know, and the extensibility was solving a problem that cannot arise.

`RuneOriginConfiguration` is a closed type with one `modelImport` field again. An unrecognised key is dropped like any other unknown property; a non-scalar value fails loudly, consistently with every other typed field. Adding a second tool means adding a field, shipped in the same version as the checker that reads it. The javadoc and docs page no longer claim otherwise, so the gap you identified closes from the other side.

One thing that needed care: a property-level `@JsonInclude` overrides the mapper default, so `origin: {}` and a null origin would have been written back as noise. Handled with an `EmptyFilter`, mirroring `RuneGeneratorsConfiguration`.

.github/scripts/annotate_generated_namespaces.py
SimonCockx: Both parts confirmed — 12 warnings + 3 notices showed 10 warnings and suppressed every notice.

The budget is now per level, so a level cannot crowd out another, and a closing notice states what did not fit:

```
::notice::2 further generated-namespace change(s) are not shown inline, because GitHub
renders at most 10 annotations per level. All 12 are listed in the workflow run summary.
```

`generated-namespaces.mdx` now says the same rather than promising an annotation per file. Tests cover the per-level split and the dropped count.

SimonCockx:
Internal links stripped from the description: the Notion design doc, the STORY reference and the private-repo PR reference are gone; `rosetta-models/demo#483` stays, since that repository is public.

I also refreshed two sections that had gone stale against the code — the marker is a closed field again, not an open map, and the accessor list now matches what actually survives.

All four threads answered. CI green: 53 Python tests, 139 rune-runtime tests, Ubuntu + Windows + docs preview.

davidalk (approved):

PR #1381, opened 2026-09-04
Format label, rule reference and transform annotations, and allow qualified names in annotations

SimonCockx:
## Formatting

`[label ...]`, `[ruleReference ...]` and `[ingest|enrich|projection ...]` were never formatted: the formatter left their whitespace exactly as written, and did not put them on a line of their own.

The cause is the shape of `RosettaFormatter`: each `format(...)` method listed the features that hold annotations (`getReferences()`, `getAnnotations()`), so the features added to the grammar later were simply not visited. `formatAnnotations` now iterates over the actual contents of an element and formats every child that is an annotation, so this class of omission does not recur when another annotation kind is added — only `isAnnotation` needs to know about it. That also picked up three gaps nobody had reported: annotations on an enum value, annotations on a `choice`, and the spacing inside an annotation qualifier (`[metadata address "pointsTo"=Bar->baz]`).

## Qualified names

`[ingest my.model.MySchema]` did not parse — `TransformAnnotation` accepted a simple name only. The same held for annotation references: `[my.annotations.myAnnotation]`. Scoping already resolves qualified names for both (the transform ref falls back to the global scope, the annotation ref uses the default scope), so only the parser rules changed.

No other cross-reference needs the same change. Every one still parsed with `ValidID` points at something resolved locally — an attribute of the enclosing type, a function input, an enum value within its enum, a schema format within the built-in `SerializationFormat` enum — where a qualified name has no meaning.

One neighbouring limitation is left alone: `schema` is a keyword that is not part of `ValidID`, so nothing can be named `schema` (a namespace segment `my.model.schema`, or an attribute). Adding it to `ValidID` leaves the parser generated from `Rosetta.xtext` unable to tell a `schema` root element from an attribute named `schema`: the build reports the ambiguity and silently disables one of the two alternatives. So it needs more than a one-line change.

## Testing

`RosettaFormatter` is covered by the 8 tests added to `RosettaFormattingTest`, one per construct changed: label and rule reference annotations on an attribute and on a choice option, transform annotations on a function, rule references in a rule source, annotation qualifiers, annotations on an enum value and on a choice, and qualified names in both annotation kinds. Each of them fails on `main` and passes here. They run with `allowUnformattedWhitespace(false)`, so they also fail if any whitespace in the model is left unformatted. The grammar change is covered by two tests in `RosettaCrossReferencingTest`.

`mvnd install` — full build, all tests, checkstyle.

https://claude.ai/code/session_01G4phbXShAy8iBNdpmrraoZ

### DIFF
--- a/rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/RosettaFormattingTest.java
+++ b/rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/RosettaFormattingTest.java
@@ -683,4 +683,162 @@ public class RosettaFormattingTest {
 					B
 				""");
 	}
+
+	@Test
+	void testLabelAnnotationOnAttribute() {
+		formatAndAssert("""
+				namespace test
+
+				type Foo:
+					bar Bar (1..1)  [  label  "My label"  ]  [ label for  baz -> qux   "Other label" ]
+				""", """
+				namespace test
+
+				type Foo:
+					bar Bar (1..1)
+						[label "My label"]
+						[label for baz -> qux "Other label"]
+				""");
+	}
+
+	@Test
+	void testRuleReferenceAnnotationOnAttribute() {
+		formatAndAssert("""
+				namespace test
+
+				type Foo:
+					bar Bar (1..1)  [ ruleReference   MyRule ]   [ruleReference  for baz ->> qux  empty]
+				""", """
+				namespace test
+
+				type Foo:
+					bar Bar (1..1)
+						[ruleReference MyRule]
+						[ruleReference for baz ->> qux empty]
+				""");
+	}
+
+	@Test
+	void testAnnotationsOnChoiceOption() {
+		formatAndAssert("""
+				namespace test
+
+				choice Foo:
+					Bar  [ label  "My label" ]
+					Qux
+				""", """
+				namespace test
+
+				choice Foo:
+					Bar
+						[label "My label"]
+					Qux
+				""");
+	}
+
+	@Test
+	void testTransformAnnotationsOnFunction() {
+		formatAndAssert("""
+				namespace test
+
+				func Foo:  [ ingest   FpML ]  [ enrich ]  [projection  XML]
+					output:
+						result Bar (1..1)
+				""", """
+				namespace test
+
+				func Foo:
+					[ingest FpML]
+					[enrich]
+					[projection XML]
+					output:
+						result Bar (1..1)
+				""");
+	}
+
+	@Test
+	void testRuleReferenceInRuleSource() {
+		formatAndAssert("""
+				namespace test
+
+				rule source TestSource {
+					Foo:
+						+ bar
+							[ ruleReference   MyRule ]
+				}
+				""", """
+				namespace test
+
+				rule source TestSource
+				{
+					Foo:
+						+ bar
+							[ruleReference MyRule]
+				}
+				""");
+	}
+
+	@Test
+	void testAnnotationRefWithQualifier() {
+		formatAndAssert("""
+				namespace test
+
+				type Foo:
+					bar Bar (1..1)  [ metadata  address   "pointsTo" = Bar -> baz ]
+				""", """
+				namespace test
+
+				type Foo:
+					bar Bar (1..1)
+						[metadata address "pointsTo"=Bar->baz]
+				""");
+	}
+
+	@Test
+	void testAnnotationsOnEnumValueAndChoice() {
+		formatAndAssert("""
+				namespace test
+
+				choice Foo:  [ deprecated ]
+					Bar
+
+				enum Baz:
+					VALUE  [ deprecated ]
+				""", """
+				namespace test
+
+				choice Foo:
+					[deprecated]
+					Bar
+
+				enum Baz:
+					VALUE
+						[deprecated]
+				""");
+	}
+
+	@Test
+	void testAnnotationsWithQualifiedNames() {
+		formatAndAssert("""
+				namespace test
+
+				type Foo:  [ test.annotations.myAnnotation ]
+					bar Bar (1..1)
+
+				func Ingest:  [ ingest  test.formats.MySchema ]
+					output:
+						result Foo (1..1)
+				""", """
+				namespace test
+
+				type Foo:
+					[test.annotations.myAnnotation]
+					bar Bar (1..1)
+
+				func Ingest:
+					[ingest test.formats.MySchema]
+					output:
+						result Foo (1..1)
+				""");
+	}
 }
--- a/rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/RosettaFormattingTest.java
+++ b/rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/RosettaFormattingTest.java
@@ -825,7 +825,7 @@ public class RosettaFormattingTest {
 				type Foo:  [ test.annotations.myAnnotation ]
 					bar Bar (1..1)
 
-				func Ingest:  [ ingest  test.formats.MySchema ]
+				func Ingest:  [ ingest  test.formats.mySchema ]
 					output:
 						result Foo (1..1)
 				""", """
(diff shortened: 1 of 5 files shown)


rune-integration-tests/src/test/java/com/regnosys/rosetta/parsing/RosettaCrossReferencingTest.java
```diff
@@ -98,4 +98,49 @@ a A (1..1)
 		modelHelper.parseRosettaWithNoIssues(model1, model2);
 		modelHelper.parseRosettaWithNoIssues(model2, model1);
 	}
+
+	@Test
+	void testFullyQualifiedSchemaCanBeUsedInTransformAnnotation() {
+		String schemaModel = """
+			namespace test.formats
+			
+			schema MySchema XML
```
SimonCockx: convention is to use lower case schema names.

rune-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.java
```diff
@@ -418,6 +431,40 @@ private void format(AnnotationRef ele, IFormattableDocument document) {
...
+	private void format(TransformAnnotation ele, IFormattableDocument document) {
+		formatSingleLineAnnotation(ele, document);
+	}
+
+	/**
+	 * Formats an annotation that always fits on one line, e.g. {@code [label for foo -> bar "My label"]}:
+	 * no space just inside the brackets, a single space between all other tokens.
+	 */
+	private void formatSingleLineAnnotation(EObject ele, IFormattableDocument document) {
+		ISemanticRegion open = regionFor(ele).keyword("[");
```
SimonCockx, line 463: `regionFor(ele).keyword("[")` returns null for `[ruleReference empty]` — that annotation gets no EObject region in the node model — and this line then throws. Xtext catches the NPE, so the annotation is silently left exactly as written:

```
type A:
    x C (0..1)  [ ruleReference   empty ]
    y C (0..1)  [ ruleReference   MyRule ]
```

Only the `MyRule` line moves onto its own line and gets normalised; the `empty` line comes out untouched. Formatting DRR's 271 files logs 916 such errors on this branch and 0 on `main` — DRR alone has 458 `[ruleReference empty]`, and two other model repos have another 263 between them.

Please null-guard `open`/`close` the way `format(Annotation)` does on line 327, and add a test that pins `[ruleReference empty]` — a guard on its own leaves the most common rule reference shape in DRR permanently unformatted.

rune-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.java
```diff
@@ -388,11 +385,27 @@ private void format(Condition ele, IFormattableDocument document) {
...
+				.forEach(annotation -> {
+					document.prepend(annotation, IHiddenRegionFormatter::newLine);
+					document.format(annotation);
+				});
+	}
+
+	private static boolean isAnnotation(EObject ele) {
+		return ele instanceof AnnotationRef
+				|| ele instanceof BuiltinAnnotation
+				|| ele instanceof RosettaDocReference
```
SimonCockx: `RosettaDocReference` extends `BuiltinAnnotationWithPath`, which extends `BuiltinAnnotation`, so this clause can never fire. Drop it.

rune-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.java
```diff
@@ -388,11 +385,27 @@ private void format(Condition ele, IFormattableDocument document) {
...
-		ele.getAnnotations().forEach(ann -> {
-			document.prepend(ann, IHiddenRegionFormatter::newLine);
-			document.format(ann);
-		});
+	/**
+	 * Formats every annotation of {@code ele} on a line of its own: annotation references such as
+	 * {@code [metadata scheme]}, doc references, labels, rule references and transform annotations.
+	 * <p>
+	 * This iterates over the actual contents of {@code ele} rather than over a fixed list of features,
+	 * so an element that gains another kind of annotation is formatted without a change to this class.
```
SimonCockx: A new *kind* of annotation still needs a clause in `isAnnotation`; what comes for free is an existing kind appearing on another element. Can we reword to say that?

rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/RosettaFormattingTest.java
```diff
@@ -683,4 +683,162 @@ void testChoice() {
...
+				""", """
+				namespace test
+
+				choice Foo:
+					[deprecated]
+					Bar
+
+				enum Baz:
+					VALUE
+						[deprecated]
```
SimonCockx, line 816: Two constructs this PR newly formats aren't pinned anywhere: an annotation on an `enum` declaration itself (`enum Foo: [deprecated]`), and a `docReference` on a `condition`. Both change real DRR output, so they are worth a test each.

SimonCockx (commented):
One defect worth fixing before merge (inline on `formatSingleLineAnnotation`); everything else is small. The switch to driving annotation formatting off `eContents()` is the right shape, and on real files it is a clear improvement.

What I checked, and how:

| Claim | How | Result |
|---|---|---|
| Effect on a real model | Ran `IFormatter2` over DRR's 271 `.rosetta` files on this branch and on `main`, diffed the two outputs | 34 files differ, and the differences are improvements: the stray whitespace-only line between a `docReference` and a `ruleReference` on an attribute is gone, and a `docReference` inside a `condition` is now indented correctly |
| `[ruleReference empty]` | Same run | 916 NPEs logged on this branch, 0 on `main` — see the inline comment |
| Churn across the estate | Text scan of CDM, DRR, FpML, ISO-20022, the demo model and three private model repos | ~31,600 label/ruleReference/transform annotations; all but 48 already sit on their own line and 357 have non-canonical inner spacing, so `rune:format` reflows hundreds of lines, not thousands |
| Qualifier and path spacing (`"pointsTo"=X->y`, `for x -> y`) | Grep of CDM and DRR | Both match what those models already write, so the differing arrow spacing between the two costs no churn |
| New parser ambiguity | `mvnd generate-sources -pl rune-lang` on both sides | `warning(200)` on `'[' 'ingest' …` already exists on `main`; this change only widens its lookahead set from `']'` to `{'.', ']'}`. No new ambiguity. |
| Downstream readers of the two cross-references | Grep of the products, components, code-generators and rune-common repos | Only `TransformAnnotation::getKind` and `Function.getAnnotations()`, both on the EMF model. Nothing reads them as text, and nothing outside `XtextResourceFormatter`'s no-parse-result fallback serialises `.rosetta` back out. |
| Robustness on broken syntax | Formatted unterminated `[label`, `[ruleReference`, `[ingest` and a bare `[` | No exception; output stable |
| Idempotency | Formatted every probe output a second time | Identical in every case |

One thing I did not verify: highlighting of the new qualified names. Reading `rune-ide/rosetta.tmLanguage.yaml`, `transformAnnotationBody` and `customAnnotationBody` match `{{identifier}}`, which contains no dot, so `[my.annotations.myAnnotation]` looks like it will colour `annotations` as an annotation attribute. `qualifiedIdentifier` is already defined in that file. Worth a look, not a blocker.

Unrelated to this PR: `typeAlias Foo: <"doc"> [deprecated]` formats to a whitespace-only line followed by the annotation, identically on `main`.

rune-integration-tests/src/test/java/com/regnosys/rosetta/parsing/RosettaCrossReferencingTest.java
SimonCockx: Renamed to `mySchema`.

rune-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.java
SimonCockx: Confirmed and fixed at the cause. `[ruleReference empty]` has only keyword assignments, so `NodeModelBasedRegionAccessBuilder.isEObjectRoot` finds nothing that marks the rule call as creating an object — no action, no cross-reference, no assigned data-type rule call — and gives it no region at all. `[ruleReference MyRule]` has the cross-reference, `[ruleReference for x -> y empty]` has the contained path, `[label "x"]` has the STRING, `[enrich]` has the enum rule call, so all of those do get one.

An explicit `{RuleReferenceAnnotation}` action in the rule gives it a region for every shape, so the annotation is formatted rather than skipped. A null guard alone would have left it permanently unformatted, as you say. Pinned by `testEmptyRuleReferenceAnnotationOnAttribute`, which reproduced the NullPointerException before the fix.

rune-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.java
SimonCockx: Right — dropped, with a comment naming the three so the next reader does not add it back.

rune-lang/src/main/java/com/regnosys/rosetta/formatting2/RosettaFormatter.java
SimonCockx: Reworded: what comes for free is an existing kind of annotation appearing on another element; a new kind still needs a clause in `isAnnotation`.

rune-integration-tests/src/test/java/com/regnosys/rosetta/formatting2/RosettaFormattingTest.java
SimonCockx: Added `testAnnotationOnEnumDeclaration` and `testDocReferenceOnCondition`. Both passed against the expected output first time, so the new behaviour on those two was already what we want.

JayasriR (approved):

PR #1382, opened 2026-09-04
Escape keywords in every part of a name

SimonCockx:
A name that collides with a keyword is written with a leading caret, e.g. `^type`, but only the first segment of a name was ever unescaped. `namespace test.^type` kept its caret in the name and crashed Java generation with `` `^type` cannot be escaped to a valid Java identifier ``; `import ^type.foo.*` never resolved; a reference into such a namespace serialized as `test.type`, which does not parse.

The root cause is that the value converters treated a name as one opaque string. Escaping is a property of a single segment, so `EscapableQualifiedNameValueConverter` now converts each segment on its own using `EscapableIDValueConverter`. Two rules that yield a name, `QualifiedNameWithWildcard` and `TypeParameterValidID`, had no converter registered at all, so Xtext passed their text through unchanged; both are registered now, and `KeywordEscapingTest` fails if the grammar grows another one.

Nothing about how a name is written is restated in Java. The escape character and the characters a name may contain both live in the `ID` terminal rule, and Xtext's own `IDValueConverter` reads them off it — it escapes with the terminal's escape character and checks what it produces by lexing it back. `EscapableIDValueConverter` delegates to one and adds only what its own grammar rule says about keywords usable as a name: `condition` in `ValidID`, `min` and `max` on top of those in `TypeParameterValidID`. That is what the deleted `ValidIDConverter` did as well, except its delegate was never given a rule, so its lexer check silently did nothing — giving it the `ID` rule is what makes the validation real, and a name that cannot be written as a Rune identifier is now rejected instead of being emitted into a document that will not parse. The separator and the wildcard come from `QualifiedName` and `QualifiedNameWithWildcard` for the same reason.

Two places write Rune source by hand rather than through the serializer, and both got the plain name wrong. The import organiser rebuilt each `import` statement from the model, so formatting a file turned `import ^type.foo.*` into `import type.foo.*`; content assist inserted `type` or `a.type` for a type named `type`. Both now go through `RosettaNameEscaper`, the one place that knows how a name is written. Neither can be handed a name it has to reject: an `import` still being typed names no namespace, and the organiser writes its keyword back rather than dropping the line the user is typing on, where escaping a null namespace would have thrown out of the whole format request and taken Sort imports and organize imports with it.

Escaping a completion before the proposal is created rather than after also closes a pre-existing gap: Xtext matches a proposal against what has been typed literally, so an escaped name could only be completed by typing its caret first. `RosettaContentProposalCreator` makes plain `ty` find `^type` too.

Whitespace is normalised in the same place for the same reason: the grammar allows `namespace foo . bar`, and the spaces were ending up in the name exactly as the caret did.

The qualified references inside annotations are worth calling out, because on `main` they look like they work. `[ruleReference my.^rule.Rule]` resolves there — not because escaping works, but because it fails symmetrically: the namespace stores its name as `my.^rule`, caret and all, and the reference text is unescaped the same wrong way, so the two match. The caret sits inside the name, which is what crashes Java generation. Break the symmetry and `main` gives up: a rule whose own name is escaped, `[ruleReference my.plain.^rule]`, is `Couldn't resolve reference to RosettaRule 'my.plain.^rule'`, because the rule's name is a plain `ValidID` and is unescaped while the reference keeps its caret. `[docReference my.plain.^body my.plain.^corpus my.plain.^segment "x"]` leaves all three unresolved. Both are tests here.

Two things were checked and left alone. The TextMate grammar already handles escapes — tokenizing namespaces, imports, qualified references, root element and attribute names, enum values, annotations, and names that collide with a section keyword (`^inputs`, `^condition`, `^to`) against the generated `rosetta.tmLanguage.json` gave identifier scopes throughout, because `identifierStart` allows a leading caret and wins the earliest-match rule against the keyword patterns. The formatter preserves escapes because it rewrites from the parsed document's own text rather than re-printing the model; `formatting-test/escapedNames.rosetta` covers that alongside the rewritten imports.

Every production file changed here is covered by a test that I confirmed fails without it: the converters, `RosettaValueConverterService` and `RosettaNameEscaper` by `KeywordEscapingTest` (ten of its twelve tests fail on `main`); `ImportManagementService` by three `ResourceFormatterServiceTest` cases; `RosettaCrossrefProposalProvider`, `RosettaContentProposalCreator`, `RosettaContentProposalProvider` and the `RosettaIdeModule` bindings by the four new `ContentAssistTest` cases. `ValidIDConverter` is deleted; the two new converters replace it. Full `mvn clean install` is green.

## Type of change

https://claude.ai/code/session_01Xx6hq9V4wnuYKh8zAtkp5f


rune-integration-tests/src/test/java/com/regnosys/rosetta/parsing/KeywordEscapingTest.java
```diff
@@ -0,0 +1,233 @@
...
+	 */
+	@Test
+	void testEveryRuleThatCanWriteANameEscapesKeywords() {
+		assertEquals(
+				Set.of("ValidID", "TypeParameterValidID", "QualifiedName", "QualifiedNameWithWildcard",
+						// Not a name: the trailing `ID` of `BigDecimal` is an exponent suffix.
+						"BigDecimal"),
+				rulesThatCanYieldAnIdentifier());
+
+		for (String rule : List.of("ValidID", "TypeParameterValidID", "QualifiedName", "QualifiedNameWithWildcard")) {
```
SimonCockx: These names are duplicated on this line.

rune-integration-tests/src/test/java/com/regnosys/rosetta/parsing/KeywordEscapingTest.java
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	@Test
+	void testCanEscapeATypeParameter() {
+		RosettaModel model = modelHelper.parseRosettaWithNoIssues("""
+				namespace test
+
+				typeAlias SmallNumber: number(^min: 1, ^max: 10)
+				""");
+		assertNoUnresolvedReferences(model);
```
SimonCockx: If the model already has no issues, what's the point of checking for unresolved references?

rune-lang/src/main/java/com/regnosys/rosetta/parsing/EscapableIDValueConverter.java
```diff
@@ -0,0 +1,96 @@
...
+ * caret, e.g. {@code ^type} - and the name it stands for, e.g. {@code type}.
+ *
+ * <p>Which keywords need a caret is derived from the grammar rule this converter belongs to:
+ * a keyword that the rule itself accepts as an identifier (e.g. {@code condition} in
+ * {@code ValidID}, or {@code min} in {@code TypeParameterValidID}) is written as is, every
+ * other keyword of the grammar is escaped. Adding a keyword to such a rule therefore needs no
+ * change here.
+ */
+public class EscapableIDValueConverter extends AbstractValueConverter<String> {
+	private static final String ESCAPE = "^";
```
SimonCockx: I have seen this character hardcoded at multiple places. Is there a canonical way to get it from the grammar? If not, is there a single logical central place where we can store it, instead of redefining it?

rune-lang/src/main/java/com/regnosys/rosetta/parsing/EscapableIDValueConverter.java
```diff
@@ -0,0 +1,96 @@
...
+	private static final Pattern IDENTIFIER = Pattern.compile("[a-zA-Z_][a-zA-Z0-9_]*");
```
SimonCockx: Does this need to be kept in sync with the grammar? Do we need to hardcode it?

website/docs/modelling-components/expressions.mdx
```diff
@@ -755,6 +755,17 @@ If conversion fails – the result is empty.
...
+
+```haskell
+namespace ^type.vehicle
+
+import ^enum.tax.*
+
+type Vehicle:
+    band ^enum.tax.VehicleTaxBandEnum (1..1)
+```
+
```
SimonCockx: No need for this documentation change. Get rid of it.

rune-integration-tests/src/test/java/com/regnosys/rosetta/parsing/KeywordEscapingTest.java
SimonCockx: Fixed — the loop takes its rules from the grammar scan now. The same names were duplicated a third time, as single-segment assertions in `testEveryNameSegmentIsEscaped` and `...Unescaped` that the scan already covers; those two tests are now only about the segments of a qualified name.

rune-integration-tests/src/test/java/com/regnosys/rosetta/parsing/KeywordEscapingTest.java
SimonCockx: No point at all — `parseRosettaWith*` already reports linking errors, which is how these tests fail on main. The helper is gone, and dropping it let every model assert no issues rather than no errors.

rune-lang/src/main/java/com/regnosys/rosetta/parsing/EscapableIDValueConverter.java
SimonCockx: Yes — it lives in the `ID` terminal rule (`terminal ID: '^'? ...`), and Xtext's own `IDValueConverter` already reads it off there. `EscapableIDValueConverter` now delegates to one bound to that rule, so the character is gone from here and from `RosettaContentProposalCreator`, which asks `RosettaNameEscaper.isEscaped` instead.

Worth knowing: the deleted `ValidIDConverter` delegated the same way, but nothing ever called `setRule` on its delegate, so the lexer check inside it silently did nothing. Giving it the `ID` rule is what makes the validation real.

rune-lang/src/main/java/com/regnosys/rosetta/parsing/EscapableIDValueConverter.java
SimonCockx: It did, and it is gone. The delegate validates by lexing back what it produced, so the character set comes from the `ID` terminal rule rather than from a copy of it here. Its messages are better than mine were, too: `only-element` now fails with "ID 'only-element' contains invalid characters: '-' (0x2d)".

The separator and the wildcard come off `QualifiedName` and `QualifiedNameWithWildcard` for the same reason, so nothing about how a name is written is restated in Java any more.

website/docs/modelling-components/expressions.mdx
SimonCockx: Dropped.

rune-lang/src/main/java/com/regnosys/rosetta/utils/ImportManagementService.java
```diff
@@ -113,17 +116,25 @@ public String toString(List<Import> imports) {
 		
 		Import previousImport = null;
 		for (Import imp : imports) {
+			String importedNamespace = imp.getImportedNamespace();
```
najihah-amran, line 119: `import foo.` (a namespace ending in the separator, or any segment that's empty) is also mid-typing, same as the null case just above — but it reaches `escapeImportedNamespace` and throws `ValueConverterException`, which is not caught anywhere in the call chain and aborts `formatCollection` for the whole batch (repro: format a doc containing just that line — throws on this branch, formats fine on `main`). Same crash from "Organize imports"/"Sort imports" quick fixes, which call this too.

SimonCockx: Agreed, and thank you for the repro — it reproduces exactly as you describe. Fixed in 9965d21ee.

Probing it turned up two more names with the same shape: `import .*` throws too, and a good import sitting beside the broken one is lost with it, since one `ValueConverterException` aborts `formatCollection` for the whole batch.

The fix is in `RosettaNameEscaper` rather than here. Its three `escape*` methods have exactly one kind of caller — code writing a name into a document that is being edited — because the parser and the serializer reach the converters directly through the package-private getters. For that caller a name that cannot be written is not an error to report, it is a name that is half typed, so the methods now give it back as it stands. The converters keep throwing, which the serializer needs: source that does not parse back is worse than an error.

That also covers the two content assist providers, so the `try`/`catch` I had written in `RosettaCrossrefProposalProvider` is gone rather than copied to a third caller. `ImportManagementService` itself is untouched by the fix. One behaviour change worth flagging: that catch used to log a warning, and no longer does.

Tests: `formatDocumentWithAHalfTypedImportedNamespace` covers your repro end to end, and `KeywordEscapingTest` now asserts both contracts against the same list of unwritable names — rejected by the converter, left alone by the escaper. Both fail without the fix.

JayasriR (approved):

PR #1386, opened 2026-09-09
Build the serializer's grammar analysis once before anything serializes

SimonCockx:
Serializing a Rune model from more than one thread on a cold injector corrupts Xtext's serializer analysis. Cold here means an injector that has not serialized anything yet, so the analysis it needs has still to be built. It surfaces as a `ConcurrentModificationException`, as `Cannot invoke "org.eclipse.xtext.util.formallang.Nfa.getStart()" because "nfa" is null`, or as `Invalid context: RosettaModel returns RosettaModel`. It is why model-import has to serialize one throwaway model up front before it can import two schema formats concurrently, and why a language server holding several languages is exposed on its first serializer request.

The unsafe state comes in two shapes, and they want different answers. Grammar-level state is the four `Map<Grammar, …>` caches in the analysis providers, the `GrammarElementDeclarationOrder` adapter that three of them install on the grammar, and the context index each `SerializationContextMap` builds on its first read. One injector has one grammar, so each of these holds exactly one entry ever: build it once and it is closed for the life of the injector.

Per-constraint state is `Constraint.getFeatures()`, `Constraint.getBody()` and `FeatureInfo.getAssignments()`, computed per constraint on the serialize path, so a model closes only the constraints it happens to reach.

An earlier version of this change guarded each provider behind a shared lock. That works. It also means enumerating Xtext's lazy fields and keeping that enumeration correct against every future version. That enumeration was already incomplete: `Constraint.getBody()` was closed only incidentally, because `getAssignments()` happens to call it for constraints that have at least one feature.

Sequencing the work closes the grammar-level shape as a class instead. `Serializer.getIContext(EObject)` is the one place every public entry point passes through before it touches a sequencer — the component that walks a model object and emits the tokens for it — and that covers every `serialize` overload, `serializeToRegions`, and `serializeReplacement`. `RosettaSerializer` overrides it to call `SerializerAnalysisWarmUp.ensureWarm()` first, so one thread builds the whole analysis while the rest wait, and nothing serializes until it is built. Call that override, and the `ensureWarm` behind it, the gate. After the first serialization the gate is a single volatile read. `serializeToRegions` matters here: it is not on `ISerializer`, so wrapping that interface would have missed it.

The serializer is bound under both `ISerializer` and the concrete `Serializer`. That is not redundancy: `XtextResourceFormatter` injects the class rather than the interface, because `serializeToRegions` is not on `ISerializer` — the same fact that makes `getIContext` the right place to gate. With the concrete class unbound it received a just-in-time instance with no gate in it, so formatting bypassed the warm-up entirely. The binding that carries this is `configureSerializer`, not the `bindISerializer()` override: Xtext already binds `ISerializer` to the concrete `Serializer`, so the interface resolves through the concrete binding either way.

`ensureWarm` sets its `warm` flag before calling `build()`, and writes the volatile `published` flag only if `build()` returns normally; a build that throws clears `warm` again so the next caller retries. That matters more than it looks: opening the gate after a failed build would put every later thread on a part-built analysis with only a log line to say so, which is a worse failure than the crash it replaces because it is silent.

Setting `warm` before the call, rather than after, is what stops the build recursing. The monitor is re-entrant: anything `build()` does that serializes arrives back at the gate on the build thread and walks straight into the synchronized block, so without the flag already set it would call `build()` again rather than block. A build with those two lines reversed fails with a `StackOverflowError`.

`build()` asks for the grammar constraints, which fills all four provider caches, then asks each of the four analyses for a single context — that fills its whole index in one pass — and then forces the per-constraint state. Two of those four indexes are read during a serialization, which is exactly where two threads would meet: the syntactic sequencer's pushdown automata at `AbstractSyntacticSequencer:337,352,442`, and the constraints at `ContextFinder:250`. The semantic-NFA map's `get` is never called in 2.38.0, and the context-type automata map is already indexed inside `getConstraints`. Both are primed anyway, as cover for a later Xtext that starts reading them.

What this does not fix is the per-constraint state, which no amount of sequencing can close, since a constraint is only reached when a model reaches it. `forceConstraintState` walks them at warm-up instead. That part is irreducible short of fixing Xtext, which wants `computeIfAbsent` on the four caches, `synchronized` on `GrammarElementDeclarationOrder.get`, and the two fields that publish a container before filling it — `getFeatures` and `getAssignments` — populating a local before assigning it instead. The same unsynchronized publication is present in Xtext 2.41.0, three minor versions on, so it is worth filing rather than only carrying a TODO.

`ENABLE_SERIALIZER_WARM_UP` stays, and no longer has anything to do with correctness. Correctness now needs no configuration at all: the gate fires on the first serialization in every process. The flag decides only whether the language server warms eagerly during construction, which it is off by default for because that slows down tests that spin up a server. Warming eagerly anywhere else is a call to `ensureWarm()` at the point of your choosing.

The case rests on the mechanism and on tests that fail deterministically. `SerializerAnalysisTest` builds a genuinely cold injector per test and asserts it is cold, rather than reusing the single injector that Xtext's `@InjectWith` test support shares across a whole run; it lives in `rune-ide`, where the injector provider is, while everything it exercises is in `rune-lang`. Each of these was checked against a mutant — a build with one line deliberately broken — by removing the thing the test protects and watching the named assertion fail: removing the gate from `getIContext`; dropping the mutual exclusion; reversing the `warm`-before-`build()` ordering; dropping the context priming for the constraints or the syntactic automata; and binding Xtext's own `Serializer` in place of `RosettaSerializer`. A separate test pins that every public serialize path still reaches the gate, since the gate is now the single point of failure. Another shares one serializer across sixteen threads, because a consumer that warms up in order to serialize concurrently is likely to hold exactly one; Xtext takes its sequencers from `Provider` fields inside each call, so the per-serialization state on `AbstractSyntacticSequencer` never spans two of them. Two tests were weak until a mutant showed it — one counted its own override rather than real gate calls, and one serialized before asserting, so it could not distinguish state the gate had built from state the serialization had built.

Worth saying plainly: the tests did not catch the unbound concrete `Serializer`, and could not have. Every one of them reaches the serializer through `ISerializer` or calls the gate directly, so a consumer that never receives a `RosettaSerializer` at all is invisible to them. It was found by asking the injector for the concrete type. The fix is visible on the artefacts themselves: build the injector, ask it for the formatter, and reflect its private `serializer` field — before, `org.eclipse.xtext.serializer.impl.Serializer`; after, `com.regnosys.rosetta.serializer.RosettaSerializer`. `bindsTheRosettaSerializerUnderBothTypes` now covers it, and a mutant confirms that assertion dies when the binding is removed.

Three lines in `build()` are kept although no test shows them protecting a serialization, and it is worth saying which.

Removing the explicit `Constraint.getBody()` call changes nothing, because every constraint in the Rosetta grammar has at least one feature and so is reached through `getAssignments()` anyway. It is kept as defence against a constraint with none, and because `getBody()` also touches `GrammarElementDeclarationOrder`.

The priming for the semantic NFAs is a third such line. Dropping it fails only the test asserting that `build()` did what `build()` says, not either concurrency test — `getSemanticSequencerNFAs` has one caller in all of 2.38.0, and it iterates the map rather than indexing it.

Removing the context priming for the context-type automata changes nothing either, and for a more specific reason: `getConstraints` forces that same index itself, when it names each constraint through `findBestConstraintName` at `GrammarConstraintProvider:564`. Asking for it directly costs one map lookup and means the gate does not depend on Xtext continuing to name constraints that way.

An end-to-end measurement against model-import corroborates it. Two cold imports of different formats start together on one process, one round per fresh JVM, configurations interleaved round by round so machine load falls on each equally. Failure rates move a long way with load, so only configurations from the same batch are comparable and nothing is pooled across batches.

| configuration | failures | rounds |
|---|---|---|
| released 10.10.1 | 120 | 150 |
| this change at `c28159844` | 0 | 150 |

and, in a second batch run the same way:

| configuration | failures | rounds |
|---|---|---|
| released 10.10.1 | 138 | 150 |
| this change at `e3f7c466e` | 0 | 150 |

p=3e-55 and p=1e-71 by Fisher's exact test. `c28159844` is this change without the later `configureSerializer` binding; `e3f7c466e` is the branch as it ships. The import path reaches the serializer through `XtextResource`'s `ISerializer` field, which both of them gate, so the second batch reruns the first rather than covering the binding — it is here because the first batch predates that commit. The control's failures are the symptoms named at the top; in the first batch, 106 of them `Nfa.getStart()` on a null NFA and the rest an invalid context.

Neither batch gives a usable figure for what the gate costs, because a failing run exits early and most of the baseline's runs failed, so its mean is not measuring the same work. A separate batch where both arms completed puts the cost of building the analysis once up front at about 0.4s per process — 6.43s against 6.01s on passing runs. The gate does that same work, so that is the order to expect, measured on the warm-up rather than on the gate itself.

The control fails 80% of rounds in the first batch and 92% in the second, and the defect is worse than either on an idle machine. Its rate fell as the first batch ran, tracking how busy the machine was, in five consecutive thirty-round windows of the same jar and the same scenario:

| rounds | control | mean run time | this change |
|---|---|---|---|
| 711-740 | 28/30 | 3.80s | 0/30 |
| 741-770 | 27/30 | 3.70s | 0/30 |
| 771-800 | 26/30 | 3.97s | 0/30 |
| 801-830 | 21/30 | 4.37s | 0/30 |
| 831-860 | 18/30 | 4.47s | 0/30 |

Run time is a proxy for contention, and it moves against the failure rate across all five. The race bites hardest when both threads genuinely run in parallel rather than being descheduled apart — the condition on a developer's laptop and a quiet CI runner rather than a loaded one. Nothing was instrumented, so that reading is a hypothesis; the rates themselves are measurements.

The consequence is that every failure rate quoted here is a lower bound rather than an estimate. Earlier batches of this experiment ran with other work competing for cores and put the same baseline as low as 0.9%, which is the defect being hidden rather than the defect being mild.

An earlier batch measured the same released DSL with model-import's existing workaround — one throwaway model serialized before any concurrency starts — at 0 failures in 150 rounds against a control failing 22. That workaround is this change done by hand in the consumer, on a DSL with no thread-safety code in it at all. The gate generalises it: it happens automatically rather than depending on a consumer remembering, and it holds a monitor, so two threads that both arrive cold block rather than relying on the warm-up having finished first.

The measurement could not have caught the failed-build case: the measurement imports well-formed fixtures, so `build()` never throws and that path never runs. A hundred and fifty green rounds would have been just as green with the gate opening on failure. That limit is not special to this one bug — it applies to every clean row here. Two well-formed fixtures on one code path show the race is gone from that path, under load that provokes it, and say nothing about paths the fixtures never reach. Nor did the measurement exercise the formatter, which does not appear on the import path at all. Correctness rests on the tests; the rates corroborate one scenario.

These counts can only see failures that throw. A dropped constraint emits different text instead, so no failure count speaks to that part either way; the argument there is the mechanism and the test.

There is independent evidence for all this in rosetta-products, written before this change existed. `ModelImporterRunnerIsolationTest` is `@Disabled` for exactly this defect — two formats importing for the first time concurrently corrupt the grammar analysis — and its javadoc records the null-NFA crash and the invalid-context one, at four failures in five runs. Re-enabling it belongs in a follow-up there rather than here. Whoever does it should not read one green run as proof. The sibling test beside it records why: unwarmed failure is roughly four in five, and the corruption persists for the life of the JVM.

`RosettaContextTypePDAProvider` is deleted. It was public, and nothing outside rune-dsl referenced it.

This will be backported to `9.x.x`.

Line references are against Xtext 2.38.0, the version this builds on.

## Type of change

https://claude.ai/code/session_01HtCUD8o1GkT2pmeCsStVPQ


rune-ide/src/test/java/com/regnosys/rosetta/serializer/RosettaContextTypePDAProviderTest.java
SimonCockx, line 1: Does this test still have benefits at all? Or is it better to delete it entirely in favour of the new serializer analysis test?

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+@Singleton
+public class SerializerAnalysisLock {
+
+	/**
+	 * Builds one step of the analysis, and hands it back ready to be read from any thread.
+	 */
+	public <T> SerializationContextMap<T> guard(Supplier<SerializationContextMap<T>> analysis) {
+		synchronized (this) {
+			SerializationContextMap<T> result = analysis.get();
```
SimonCockx: This forces the context index but not the `Constraint` values behind it. `Constraint.getFeatures()` assigns its array at `GrammarConstraintProvider:267`, then runs `computeLowerBounds`/`computeUpperBounds`, and only fills the slots at `:272` — so for the length of those two computations the field holds a non-null array of nulls. A second thread needs no reordering to see it, just a plain interleaving, and `ContextFinder:195` reads `getFeatures()[refID] == null` and drops a constraint that is actually valid. That branch is unreachable single-threaded, so the failure is usually silent: a different context is chosen and different text is emitted, which is why a failure-count run would not see it. `FeatureInfo.getAssignments():477` publishes its list before filling it the same way.

Walk `values()` here and force `getFeatures()`, and `getAssignments()` on each feature, while still holding the monitor.

Read from the 2.38.0 sources, not reproduced at runtime.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,95 @@
...
+	 * the entry point for a standalone consumer that is about to serialise from several threads.
+	 *
+	 * <p>It asks only for the grammar constraints. Everything the chain reaches through them, down to
+	 * each step's context index, is made ready to read by the guarded call itself rather than by
+	 * anything here, so a consumer that skips this warm-up is slower on its first serialization but
+	 * no less safe.
+	 */
+	public void warmUp() {
+		long started = System.nanoTime();
+		constraintProvider.getConstraints(grammarAccess.getGrammar());
```
SimonCockx: `getConstraints` does not force the lazy `Constraint` values, whereas model-import's current workaround serializes a real model and so fills whatever the hot path touches. As written, swapping that workaround for this call trades protection it has today for coverage of three more providers.

Forcing those values under the lock fixes both at once: this then dominates the workaround on providers and on constraints, and the swap becomes safe.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	@Test
+	void eachInjectorGetsItsOwnLockAndGrammar() {
+		Injector other = new RosettaIdeInjectorProvider().getInjector();
+
+		// This is what makes a per-injector lock enough. Everything it has to protect — the four
+		// providers, and the grammar they attach their analysis to — is scoped to one injector, so a
+		// second injector has nothing to race on.
+		assertNotSame(injector.getInstance(SerializerAnalysisLock.class),
```
SimonCockx: This pins that two injectors get different locks, but nothing pins that the four providers within one injector share the same lock — which is the argument the whole design rests on. Removing `@Singleton` from `SerializerAnalysisLock` gives each provider its own lock and leaves the entire suite green; I ran that mutant to check.

Assert the four providers hold the same `SerializerAnalysisLock` instance.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+import org.eclipse.xtext.serializer.analysis.SerializationContextMap;
+
+import jakarta.inject.Singleton;
+
+/**
+ * The single monitor behind which this language's serializer analysis is built.
+ *
+ * <p>Xtext's four analysis providers each cache their result per grammar in a plain
+ * {@link java.util.HashMap} with an unsynchronized check-then-put. One monitor per provider is not
+ * enough to fix that: all four also reach
```
SimonCockx: Three of the four, not all four. Only `GrammarConstraintProvider:245,623`, `SemanticSequencerNfaProvider:275` and `SyntacticSequencerPDAProvider:460` reach it; `ContextTypePDAProvider` has no sort or comparator path, and nor does anything it reaches. It still belongs on the shared lock for its own cache, so the design is unaffected — but this sentence is the argument for the design, so it is worth being exact.

rune-lang/src/main/java/com/regnosys/rosetta/RosettaRuntimeModule.java
```diff
@@ -136,11 +145,34 @@ public Class<? extends ResourceFormatterService> bindResourceFormatterService()
 		return XtextResourceFormatter.class;
 	}
 
+	// Xtext's serializer analysis is not thread safe on a cold grammar: each of its four providers
+	// caches its result per grammar in a plain HashMap with an unsynchronized check-then-put, and all
+	// four also mutate the shared GrammarElementDeclarationOrder. The Rosetta subclasses below build
```
SimonCockx: Same "all four" as in `SerializerAnalysisLock` — it is three of them.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+ * {@link SerializerAnalysisWarmUp} is for: warm up and no request ever waits behind the analysis.
+ * Once the analysis is built each guarded call is a hash lookup.
+ *
+ * TODO: contribute thread safety for this analysis to Xtext, and drop the lock.
+ */
+@Singleton
+public class SerializerAnalysisLock {
+
+	/**
+	 * Builds one step of the analysis, and hands it back ready to be read from any thread.
```
SimonCockx: "ready to be read from any thread" holds for the `SerializationContextMap` index, not for the object graph it points at. Worth scoping to the index unless the constraint values get forced too.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+ * The single monitor behind which this language's serializer analysis is built.
+ *
+ * <p>Xtext's four analysis providers each cache their result per grammar in a plain
+ * {@link java.util.HashMap} with an unsynchronized check-then-put. One monitor per provider is not
+ * enough to fix that: all four also reach
+ * {@link org.eclipse.xtext.serializer.analysis.GrammarElementDeclarationOrder#get}, which caches by
+ * adding an adapter to the grammar, iterating the adapter lists of the used grammars while removing
+ * from them. Two threads inside two <em>different</em> providers therefore still corrupt that shared
+ * state — in practice with a {@link java.util.ConcurrentModificationException}. One lock across the
+ * whole analysis is what makes the chain safe.
```
SimonCockx: "makes the chain safe" — say "makes these four steps safe". Same for "Five lazily-filled caches sit on that path" in the description; both read as exhaustive.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+	/**
+	 * Forces the lazy context index that {@link SerializationContextMap} keeps in an unsynchronized
+	 * field of its own and fills in full on the first {@link SerializationContextMap#get} call.
+	 *
+	 * <p>Guarding only the four providers leaves that first {@code get} to happen on the serializer
+	 * hot path, on whichever thread reaches it — a second check-then-fill just past this lock, where
+	 * two threads on a cold injector both see a null index and one reads the map the other is still
+	 * filling. It surfaces one frame further out than the provider caches do, as
+	 * {@code Invalid context: RosettaModel returns RosettaModel} from the semantic sequencer or a
+	 * bare {@link IllegalStateException} from the syntactic one.
```
SimonCockx: The syntactic side throws an NPE, not an `IllegalStateException` — `AbstractSyntacticSequencer.init:442` uses `Preconditions.checkNotNull(state, "Invalid context: " + context)`. The bare `IllegalStateException`s elsewhere in that class are traversal failures.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+ * state — in practice with a {@link java.util.ConcurrentModificationException}. One lock across the
+ * whole analysis is what makes the chain safe.
+ *
+ * <p>The lock belongs to one language's injector, along with the providers it guards and the
+ * grammar they attach their analysis to, so two languages in one language server never wait on
+ * each other. Within a language the first serialization does block the second, which is what
+ * {@link SerializerAnalysisWarmUp} is for: warm up and no request ever waits behind the analysis.
+ * Once the analysis is built each guarded call is a hash lookup.
+ *
+ * TODO: contribute thread safety for this analysis to Xtext, and drop the lock.
```
SimonCockx: Give this an issue link and move it out of the javadoc block, where it renders into the generated docs. The previous TODO named the fix — `ConcurrentHashMap.computeIfAbsent` — which is worth keeping, and given the lazy state above, upstreaming really is the only complete fix rather than a nicety.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.eclipse.xtext.serializer.ISerializationContext;
+import org.eclipse.xtext.serializer.analysis.SerializationContextMap;
+
+import jakarta.inject.Singleton;
+
+/**
+ * The single monitor behind which this language's serializer analysis is built.
+ *
+ * <p>Xtext's four analysis providers each cache their result per grammar in a plain
```
SimonCockx: This paragraph appears seven times across the PR — here, in each of the four providers, in `RosettaRuntimeModule`, and in the test. The wrong "all four" is that duplication already costing something: one sentence, corrected in one place, still wrong in the other. Keep the explanation here and leave each provider with its one-line "builds X under the lock".

rune-ide/src/main/java/com/regnosys/rosetta/ide/serializer/SerializerWarmUpService.java
```diff
@@ -19,29 +19,29 @@
...
+import com.regnosys.rosetta.serializer.SerializerAnalysisWarmUp;
 import com.regnosys.rosetta.utils.EnvironmentUtil;
 
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 
 /**
- * Triggers a non-eager, asynchronous warm-up of the serializer
- * context-type PDA cache for every registered Xtext language.
+ * Triggers a non-eager, asynchronous {@link SerializerAnalysisWarmUp} for every
```
SimonCockx: Not every registered language — `warmUp` warms only those binding the Rosetta providers, as the comment at line 94 says. Add "that binds the Rosetta analysis providers".

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,95 @@
...
+
+	@Inject
+	public SerializerAnalysisWarmUp(IGrammarAccess grammarAccess, IGrammarConstraintProvider constraintProvider) {
+		this.grammarAccess = grammarAccess;
+		this.constraintProvider = constraintProvider;
+	}
+
+	/**
+	 * Builds the analysis on the calling thread, returning once every provider is populated. This is
+	 * the entry point for a standalone consumer that is about to serialise from several threads.
```
SimonCockx: Worth saying "one `ISerializer` per thread" here. `AbstractSyntacticSequencer:98-100` keeps per-serialization mutable state and `AbstractRosettaRuntimeModule:95-106` binds it unscoped, so a fresh instance per thread is correct — but this sentence invites exactly the consumer who would cache one in a field and call it from several threads.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/RosettaSyntacticSequencerPDAProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+@Singleton
+public class RosettaSyntacticSequencerPDAProvider extends SyntacticSequencerPDAProvider {
+	@Inject
+	private SerializerAnalysisLock lock;
+
+	@Override
+	public SerializationContextMap<ISynAbsorberState> getSyntacticSequencerPDAs(Grammar grammar) {
+		return lock.guard(() -> super.getSyntacticSequencerPDAs(grammar));
+	}
```
SimonCockx: Worth a note that `SyntacticSequencerPDAProvider:460` reaches `GrammarElementDeclarationOrder.get` from `getAmbiguousSyntax()`, outside this lock, and is safe only because `GrammarConstraintProvider:623` installs that adapter under the lock first. A consumer going straight to the syntactic PDAs would get the unguarded `eAdapters()` mutation back. Nothing currently records that ordering.

rune-lang/src/main/java/com/regnosys/rosetta/RosettaRuntimeModule.java
```diff
@@ -136,11 +145,34 @@ public Class<? extends ResourceFormatterService> bindResourceFormatterService()
...
+	// the analysis behind one SerializerAnalysisLock; see that class for why one lock per provider is
+	// not enough. Each is bound under both its interface and the concrete Xtext class, because an
+	// unbound concrete class would get a just-in-time instance with a second, unguarded cache of its
+	// own — which is what SyntacticSequencerPDAProvider would otherwise end up with, as it injects the
+	// concrete ContextTypePDAProvider rather than the interface.
```
SimonCockx: Keep the four bindings, but this is true of one of them today: `SyntacticSequencerPDAProvider:527-528` is the only concrete-class injection anywhere in Xtext 2.38.0's serializer. Worth saying the other three are defensive against a future Xtext adding one, since otherwise the next reader has to grep to find out which.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,86 @@
...
+ * TODO: contribute thread safety for this analysis to Xtext, and drop the lock.
+ */
+@Singleton
+public class SerializerAnalysisLock {
+
+	/**
+	 * Builds one step of the analysis, and hands it back ready to be read from any thread.
+	 */
+	public <T> SerializationContextMap<T> guard(Supplier<SerializationContextMap<T>> analysis) {
+		synchronized (this) {
```
SimonCockx: Have you measured the steady-state cost? `BacktrackingSemanticSequencer:476` and `ContextFinder:267` call `getConstraints` per `EObject`, so every serialized object takes this monitor and concurrent serialization within a language fully serialises on it — on the path this PR exists to open up. A volatile double-checked read would keep the warm case off the monitor. No change needed if you have already looked at it.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
```diff
@@ -0,0 +1,263 @@
...
+		return (Map<Grammar, ?>) cache.get(injector.getInstance(boundInterface));
+	}
+
+	private static class BlockingWarmUp extends SerializerAnalysisWarmUp {
+		private final AtomicInteger runs = new AtomicInteger();
+		private final CountDownLatch entered = new CountDownLatch(1);
+		private final CountDownLatch release = new CountDownLatch(1);
+
+		BlockingWarmUp() {
+			super(null, null);
```
SimonCockx: One line on why the nulls are safe here — that `warmUp()` is overridden and never touches the injected fields.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
```diff
@@ -0,0 +1,263 @@
...
+		assertTrue(cache(ISemanticSequencerNfaProvider.class, SemanticSequencerNfaProvider.class).containsKey(grammar));
+		assertTrue(cache(ISyntacticSequencerPDAProvider.class, SyntacticSequencerPDAProvider.class)
+				.containsKey(grammar));
+		assertTrue(cache(IContextTypePDAProvider.class, ContextTypePDAProvider.class).containsKey(grammar));
+
+		// The warm-up asks only for the constraints, so that every step it reaches through them comes
+		// back with its context index built is a property of the guarded call rather than of the
+		// warm-up. Checked here rather than assumed, because a consumer that warms up and then
+		// serialises from several threads is relying on it.
+		assertContextIndexBuilt("grammar constraints",
```
SimonCockx: JUnit stops at the first failed assertion, so on every mutant I ran only the first of these four fired and the other three were never exercised. `assertAll` would prove all four, and would tell you which step regressed rather than just the first.

SimonCockx (commented):
The lock, the four subclasses and the double binding are right, and the description made it easy to check rather than take on trust. I re-derived the argument against the Xtext 2.38.0 sources rather than reading it back to you:

|---|---|---|
| Four caches on the path | `Map<Grammar,…>` across all six analysis providers | Holds — `ContextPDAProvider`/`GrammarPDAProvider` cache nothing |
| Guarded entry points complete | every public/protected member of the four Xtext classes | Holds — one cache-touching method each, all overridden |
| One `get()` fills the whole index | `SerializationContextMap:121-131` | Holds — asking for one context is enough |
| One shared lock, not four | `GrammarElementDeclarationOrder:36-51` | Holds; three providers reach it, not four |
| Double binding needed | `SyntacticSequencerPDAProvider:528` injects the concrete class | Holds — and it is the only such injection in Xtext 2.38.0 |
| Per-injector lock is enough | `GrammarProvider:62` builds its own `XtextResourceSet` | Holds — two injectors share no grammar object |
| `RuleNames` adapter also covered | `RuleNames:91-100`, sole caller `GrammarPDAProvider:179` | Holds — inside the guard |
| Analysis ready to read on release | `GrammarConstraintProvider:267-272`, `:477-486` | **Does not hold** — see the comment on `guard` |

Two things to settle. The analysis handed back is still not fully built: `Constraint.getFeatures()` publishes its array before filling it, and `ContextFinder` reads it on the ordinary `serialize(EObject)` path. That is the same shape as the context-index race you fixed one frame in, with a wider window, and it fails silently — which is why the 150-round table reads zero for it either way. And the model-import swap the description describes would, as things stand, cost model-import protection it has today. Forcing the constraint values under the lock resolves both: `warmUp()` then dominates the throwaway-serialization workaround on providers and on constraints.

Worth being deliberate about where that stops. Forcing every lazy field is unbounded — the syntactic PDA has six more, and a future Xtext can add one with nothing to make it fail loudly. Forcing the constraint map is bounded, sits on the plain `serialize` path, and is about ten lines; the syntactic PDA's lazy state is idempotent over immutable values and I would leave it. That makes the TODO to upstream this to Xtext the only complete fix rather than a nicety, which is worth saying in the comment itself.

Two asks outside the diff. Does this want backporting to `9.x.x`? It carries only the earlier partial fix — one guarded provider, no shared lock, no context-index priming — which is weaker than the second row of your own table, and bundle 11.x runs on it. And naming the model-import follow-up in the description would keep that debt visible rather than parked in a scratch branch.

Two small things I could not anchor inline. `RosettaContextTypePDAProviderTest.bindsRosettaContextTypePDAProvider` is now a strict subset of `bindsTheRosettaProviderForEveryCachedStep`; `concurrentCallsWaitForFirstComputationAndReuseCache` is not redundant though — it is the only test asserting mutual exclusion actually happens, which `assertSame` alone cannot catch — so moving that one across and dropping the rest of the file would work. And in the description: the sixteen threads receive four objects, not one, since line 208 asserts within each group of four; "covers four things" understates a six-test class; and it is worth saying the test lives in `rune-ide` while everything it tests is in `rune-lang`.

Verified by reading sources and by running mutants against the suite; the concurrency arguments are static reachability, not a reproduced failure, and I have flagged which is which inline.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: Correcting this — I wrote it off an interim result and it is wrong in your favour. Three of the four assertions are load-bearing: the constraint, semantic-NFA and syntactic-PDA ones are each killed by unguarding their own provider.

The one that proves nothing is the context type PDAs. `GrammarConstraintProvider:635` fetches the type PDAs and `:638` reaches `findBestConstraintName`, which does `nfaUtil.collect(typePDAs.get(ctx))` at `:564` — that `get` builds `SerializationContextMap.keys` as a side effect while the constraints are being built. So the assertion passes whether or not `RosettaContextTypePDAProvider` is guarded, which is what I measured: unguarding it leaves both these tests green.

To give that assertion teeth it has to run against the context type PDAs on a cold injector before anything asks for the constraints. `assertAll` is still worth having, but it was not the real point here.

SimonCockx: Numbers now in, since I quoted this mutant without one: splitting the lock four ways survived 5 of 5 clean runs, every run `Tests run: 8, Failures: 0`.

Worth saying the suite is not simply blind to lock changes — removing mutual exclusion while keeping `indexByContext` was killed 3 of 3 by both concurrency tests. It catches the lock going missing; it does not catch the lock being split.

The same test also passes with the monitor made process-global (`synchronized (SerializerAnalysisLock.class)`), because `eachInjectorGetsItsOwnLockAndGrammar` only asserts the two instances differ, which a static monitor satisfies too. So the scoping is unpinned in both directions: nothing fails if the lock becomes narrower than one per injector, and nothing fails if it becomes wider.

A test driving two *different* providers concurrently on a cold injector would close the first; asserting the two injectors' providers hold their own injector's lock instance would close the second.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
```diff
@@ -0,0 +1,263 @@
...
+	private Injector injector;
+
+	@BeforeEach
+	void createColdInjector() {
+		// A fresh injector provider builds a fresh injector, so every cache in the chain starts empty.
+		// The injector the InjectionExtension hands out is shared across the whole test run and is
+		// warm by the time this class runs.
+		injectorProvider = new RosettaIdeInjectorProvider();
+		injectorProvider.setupRegistry();
+		injector = injectorProvider.getInjector();
```
SimonCockx, line 102: The comment above explains why this injector has to be cold, but nothing enforces it, and the whole power of `concurrentFirstUseGivesEveryThreadTheSameAnalysis` rests on it — once the caches are warm every thread returns the same object without entering the build. I swapped in a shared warm injector and removed `synchronized`, and that test passed.

The counter-example is adjacent rather than hypothetical: `RosettaContextTypePDAProviderTest`, in this package and edited by this PR, still uses `@InjectWith(RosettaIdeInjectorProvider.class)`, which shares one injector across the run. Anyone harmonising the two files would reach for it and silently defang this one.

Four lines using the `cache(...)` helper you already have — assert the four caches are empty here — makes that fail loudly instead.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
```diff
@@ -0,0 +1,263 @@
...
+				injector.getInstance(ISemanticSequencerNfaProvider.class).getSemanticSequencerNFAs(grammar));
+		assertContextIndexBuilt("syntactic sequencer PDAs",
+				injector.getInstance(ISyntacticSequencerPDAProvider.class).getSyntacticSequencerPDAs(grammar));
+		assertContextIndexBuilt("context type PDAs",
+				injector.getInstance(IContextTypePDAProvider.class).getContextTypePDAs(grammar));
+	}
+
+	@Test
+	void warmUpStartsOncePerInjector() throws Exception {
+		BlockingWarmUp warmUp = new BlockingWarmUp();
```
SimonCockx, line 308: Building this by hand means the test proves `warmUpAsync` memoises per instance, but not the "once per injector" its name claims — dropping `@Singleton` from `SerializerAnalysisWarmUp` leaves it green. One `assertSame` on two `injector.getInstance(SerializerAnalysisWarmUp.class)` calls would cover the other half.

rune-ide/src/main/java/com/regnosys/rosetta/ide/serializer/SerializerWarmUpService.java
```diff
@@ -91,10 +91,10 @@ public void warmUp() {
...
 
 	private void warmUp(IResourceServiceProvider provider) {
-		IContextTypePDAProvider pdaProvider = provider.get(IContextTypePDAProvider.class);
-		IGrammarAccess grammarAccess = provider.get(IGrammarAccess.class);
-		if (pdaProvider instanceof RosettaContextTypePDAProvider && grammarAccess != null) {
-			((RosettaContextTypePDAProvider) pdaProvider).warmUpAsync(grammarAccess.getGrammar());
+		// Only languages that bind the Rosetta analysis providers are warmed up: for any other
+		// language the warm-up would build an analysis nothing here is going to serialise.
+		if (provider.get(IGrammarConstraintProvider.class) instanceof RosettaGrammarConstraintProvider) {
+			provider.get(SerializerAnalysisWarmUp.class).warmUpAsync();
```
SimonCockx, line 83: This class has no test of its own — its only reference is `RosettaLanguageServerImpl`. The rewiring here is the part that decides whether the language server warms anything at all, and both branches of the new `instanceof` filter, and the flag check above it, are currently unexercised.

SimonCockx (commented):
Three more test gaps, from finishing the mutation run I mentioned. Same theme as my earlier comment on `eachInjectorGetsItsOwnLockAndGrammar`, so nothing here changes the verdict — the tests pin the behaviour well and the structure the argument rests on less so.

Worth stating what the suite does catch, because it is most of it: all four providers unguarded individually, all four at once, the `indexByContext` line deleted (killed deterministically by exactly the two tests you named), mutual exclusion removed, a no-op warm-up, and the concrete `ContextTypePDAProvider` binding removed. `assertContextIndexBuilt` is not tautological either — `values()` does not build the index, only `get()` does.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
```diff
@@ -0,0 +1,180 @@
...
+	 *
+	 * <p>{@link IConstraint#getFeatures} is documented as leaving a null slot for a feature with no
+	 * assignment. In 2.38.0 the one implementation fills every slot, so this loop should not meet a
+	 * null — but the contract allows one, and {@code ContextFinder} reads the array expecting one, so
+	 * this skips rather than throws. Forcing is an optimisation of when work happens; it should not
+	 * be the thing that turns a tolerated null into a crash.
+	 */
+	private static void fillConstraintValues(SerializationContextMap<IConstraint> constraints) {
+		for (SerializationContextMap.Entry<IConstraint> entry : constraints.values()) {
+			for (IFeatureInfo feature : entry.getValue().getFeatures()) {
```
SimonCockx: `Constraint.getBody()` is a third lazy field on this path, closed here only because `getAssignments()` happens to call it, and only for constraints that have at least one feature.

SimonCockx (commented):
Taking the approach first, since it decides whether the rest matters. The concern is that going through Xtext's classes one by one may already be incomplete and is hard to keep complete. I think that is right, and that the enumeration is avoidable.

The unsafety is two hazards with different shapes. Grammar-level state is the four `Map<Grammar, …>` caches (`ContextTypePDAProvider:226`, `SemanticSequencerNfaProvider:202`, `SyntacticSequencerPDAProvider:524`, `GrammarConstraintProvider:516`) and the `GrammarElementDeclarationOrder` adapter. One injector has one grammar, so each holds exactly one entry ever, and any call that reaches a provider once closes its cache permanently. `SerializationContextMap.get` belongs here too rather than with the per-context state: the first `get()` indexes every context in one pass. Per-constraint state is `Constraint.getFeatures()` (`GrammarConstraintProvider:261`), `FeatureInfo.getAssignments()` (`:456`) and `Constraint.getBody()`, all computed per constraint on the serialize path, so a model closes only the constraints it reaches.

Separating those answers both options. The enumeration is only needed because this guards access to each provider instead of sequencing it. And a sample model is a sound answer to the grammar-level hazard that does not need to trigger every route, because there is one entry per cache to close — it just cannot close the per-constraint one, which is what makes it feel like it has to be exhaustive.

Sequencing instead: `Serializer.getIContext(EObject)` is a funnel every public path reaches before touching a sequencer, including `serializeToRegions`, which is not on `ISerializer`.

```java
public class RosettaSerializer extends Serializer {
    @Inject private SerializerAnalysisWarmUp warmUp;

    @Override
    protected ISerializationContext getIContext(EObject semanticObject) {
        warmUp.ensureWarm();
        return super.getIContext(semanticObject);
    }
}
```

```java
public void ensureWarm() {
    if (published) return;          // volatile
    synchronized (this) {
        if (warm) return;           // re-entrant: the sample serialization passes through
        warm = true;
        try { build(); } finally { published = true; }
    }
}

private void build() {
    SerializationContextMap<IConstraint> constraints =
            constraintProvider.getConstraints(grammarAccess.getGrammar());  // fills all four caches
    forceConstraintState(constraints);
    serializeSampleModel();
}
```

Setting `warm` before `build()` is what lets the sample serialization re-enter and terminate; other threads still block on the monitor, and `published` is written only after `build()` returns. After warm-up the gate is one volatile read, so threads run free.

That is one `bindISerializer` in place of the eight bindings at `RosettaRuntimeModule:139-176`, and it deletes the four provider subclasses, `SerializerAnalysisLock`, the concrete-class workaround for `SyntacticSequencerPDAProvider:528`, the `ENABLE_SERIALIZER_WARM_UP` flag, and model-import's throwaway serialization. It also closes the ordering hazard recorded on `RosettaSyntacticSequencerPDAProvider` by construction, since nothing serializes before warm-up. If Xtext adds a grammar-level lazy field later, the gate closes it with no code change; a new per-constraint one leaves us where this PR is, with the sixteen-thread test as the net either way.

What survives is `fillConstraintValues`, and that is irreducible short of fixing Xtext: four caches to `computeIfAbsent`, build-into-a-local-then-assign in `SerializationContextMap.get`, `Constraint.getFeatures` and `FeatureInfo.getAssignments`, and `synchronized` on `GrammarElementDeclarationOrder.get`. Worth filing upstream rather than only carrying the TODO, since it deletes the gate as well.

Two things I have not checked. Whether anything downstream — rune-common, rosetta-code-generators, model-import — injects `ContextFinder`, `ISemanticSequencer` or the providers directly and would bypass the gate; nothing in rune-dsl's main sources does. And the sample model must not be able to fail the warm-up, since `getConstraints` has done the load-bearing part by the time it runs.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/RosettaContextTypePDAProviderTest.java
SimonCockx: Deleted. The mutual-exclusion assertion is not lost: dropping `synchronized` from `ensureWarm` is caught by the sixteen-thread test.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Done. `forceConstraintState` runs on the build thread inside the gate. Sequencing cannot close per-constraint state, so this is the one hazard kept from the previous design.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Right on both counts. The gate forces these before anything serializes, so there is no protection left to trade away.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: The lock is gone. Removing the gate, dropping the mutual exclusion, and reversing the `warm`-before-`build()` ordering each now fail a named test; all three were run as mutants.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Correct, and the sentence went with the class. The gate does not depend on which providers reach it, because nothing serializes until the build thread has installed the adapter.

rune-lang/src/main/java/com/regnosys/rosetta/RosettaRuntimeModule.java
SimonCockx: Gone. The eight bindings are replaced by one `bindISerializer`.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Gone. The new javadoc separates grammar-level state from per-constraint state, so no unqualified claim is left to scope.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Gone. The replacement says what it closes and what it does not.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: The wrong sentence is deleted rather than corrected, and nothing in the new code restates it.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Moved out of the javadoc into a body comment, naming the six upstream fixes. Still no link: no upstream issue exists yet.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Four of those seven files are deleted. The explanation now lives in `SerializerAnalysisWarmUp` only.

rune-ide/src/main/java/com/regnosys/rosetta/ide/serializer/SerializerWarmUpService.java
SimonCockx: Fixed.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/RosettaSyntacticSequencerPDAProvider.java
SimonCockx: Closed by construction. `build()` installs the adapter before anything serializes, so that read is a cache hit and no consumer can reach it first.

rune-lang/src/main/java/com/regnosys/rosetta/RosettaRuntimeModule.java
SimonCockx: All eight bindings are gone, replaced by one `bindISerializer`. There is no concrete-class workaround left to explain.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Measured, and the gate dissolves the question: the warm path is a volatile read rather than a monitor, so nothing is taken per serialized object.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: Added.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: Done, and you were right that three of the four were load-bearing. `assertAll` throughout, including the cold-injector check.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: Fixed. `assertColdInjector()` asserts all four caches are empty rather than relying on it.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: Fixed. It now asserts on two `injector.getInstance(...)` calls.

rune-ide/src/main/java/com/regnosys/rosetta/ide/serializer/SerializerWarmUpService.java
SimonCockx: Added `SerializerWarmUpServiceTest`.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisLock.java
SimonCockx: Now forced explicitly. No mutant kills it, though: every Rosetta constraint has at least one feature, so `getAssignments()` reaches it anyway. Kept as defence against a constraint with none, and because it also touches `GrammarElementDeclarationOrder`.

rune-lang/src/main/java/com/regnosys/rosetta/RosettaRuntimeModule.java
SimonCockx: Correcting my reply above: there is still a double binding, and for a better reason than before. The serializer is bound under both `ISerializer` and the concrete `Serializer`, because `XtextResourceFormatter` injects the class to reach `serializeToRegions`, which the interface does not declare. Leaving the concrete class unbound gave it a just-in-time ungated instance, so formatting bypassed the gate entirely — fixed in `3c5168e9d`. Found by reading the built injector's bindings, not by a test.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Pushing back on this one, and leaving the thread open so you can weigh in. I do not think "one `ISerializer` per thread" is right. `Serializer.serialize` obtains all three sequencers from providers on every call — `Serializer:101-103` and `:114-116` — and `AbstractRosettaRuntimeModule:95-102` binds them unscoped, so the per-serialization state in `AbstractSyntacticSequencer.contexts` lives on an instance created per serialization rather than per serializer. That is what makes a shared serializer safe, and it is why `XtextResourceFormatter` can hold one in a field. The consumer who must not share is one caching a sequencer. Happy to add a note saying that instead, if you agree it is the more useful thing to warn about.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,228 @@
...
+			if (warm) {
+				return;
+			}
+			// Set before build(), not after: anything build() does that serializes re-enters this
+			// method on this thread, and would otherwise block on a monitor it already holds.
+			warm = true;
+			try {
+				build();
+			} finally {
+				published = true;
```
SimonCockx: Set `published` only when `build()` returns normally. A throwing build opens the gate for good: every later thread skips the monitor and serializes on the half-built analysis this class exists to prevent, and `runWarmUp` logs rather than propagates, so the async path fails silently. It also contradicts the field's own javadoc at line 92 — "knows the analysis is finished rather than merely started" is exactly what a thread cannot conclude from a `finally`. Resetting `warm = false` on the way out lets the next caller retry.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,228 @@
...
+ * <p>Grammar-level state is the four per-grammar caches in the analysis providers, the shared
+ * {@code GrammarElementDeclarationOrder} adapter on the grammar, and the context index inside each
+ * {@link SerializationContextMap}. An injector has exactly one grammar, so each of these holds one
+ * entry ever and any call that reaches it once closes it permanently. {@link #build()} closes all of
+ * them by asking for the grammar constraints, which pull the other three providers in dependency
+ * order.
+ *
+ * <p>Per-constraint state is {@code Constraint.getFeatures()}, {@code Constraint.getBody()} and
+ * {@code FeatureInfo.getAssignments()}, computed per constraint on the serialize path, so a model
+ * closes only the constraints it happens to reach. Each of the three assigns its container to the
```
SimonCockx: `getBody()` computes into a local and assigns last (`GrammarConstraintProvider:246-250`), so it is not one of the assign-then-fill fields — only `getFeatures` and `getAssignments` are, and the dropped-constraint symptom described next belongs to `getFeatures` alone. Forcing the body is still right, for the other reason, but the enumeration is what the next reader will trust. Same sentence in the commit message.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,228 @@
...
+
+	/**
+	 * Asking for the grammar constraints fills all four provider caches: the constraints need the
+	 * semantic sequencer NFAs, which need the syntactic sequencer PDAs, which need the context type
+	 * PDAs. It also installs the {@code GrammarElementDeclarationOrder} adapter on the grammar, so the
+	 * later reads of it from the serialize path are cache hits that mutate nothing.
+	 *
+	 * <p>Each of the four analyses is then asked for one context. Filling the provider caches is not
+	 * the same as finishing the analyses: a {@link SerializationContextMap} indexes its contents on
+	 * the first {@code get()}, and for three of the four that first {@code get()} would otherwise
```
SimonCockx: Two of the four, not three. `.get()` on the semantic-NFA map is never called anywhere in Xtext 2.38.0 — `getSemanticSequencerNFAs` has one caller, `GrammarConstraintProvider:624`, which iterates `values()` — and the context-type map is already indexed inside `getConstraints` itself at `:564`. On the serialize path it is the syntactic PDAs (`AbstractSyntacticSequencer:337,352,442`) and the constraints (`ContextFinder:250`). Indexing all four is still worth doing as cover for a later Xtext; worth saying that instead.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,228 @@
...
+		indexByContext(contextTypePdaProvider.getContextTypePDAs(grammar));
+		indexByContext(syntacticPdaProvider.getSyntacticSequencerPDAs(grammar));
+		indexByContext(nfaProvider.getSemanticSequencerNFAs(grammar));
+		SerializationContextMap<IConstraint> constraints = constraintProvider.getConstraints(grammar);
+		indexByContext(constraints);
+		forceConstraintState(constraints);
+	}
+
+	/**
+	 * Forces every constraint's body, features and assignments, and the context index of the map
```
SimonCockx: Drop "and the context index of the map holding them" — `indexByContext` does that, separately, at line 187.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,228 @@
...
+
+	private final IGrammarAccess grammarAccess;
+	private final IGrammarConstraintProvider constraintProvider;
+	private final ISemanticSequencerNfaProvider nfaProvider;
+	private final ISyntacticSequencerPDAProvider syntacticPdaProvider;
+	private final IContextTypePDAProvider contextTypePdaProvider;
+
+	/**
+	 * Set before {@link #build()} runs, and read only under the monitor. This is what makes the gate
+	 * re-entrant: anything {@code build()} itself does that serializes would otherwise deadlock on a
```
SimonCockx: It is a `StackOverflowError`, not a deadlock. `synchronized` is reentrant, so the thread re-acquires the monitor, sees `warm` still false and recurses into `build()` forever. I ran the mutation: moving `warm = true` after `build()` fails `theGateLetsTheBuildItselfSerialize` in 11s with `java.lang.StackOverflowError`, no timeout. Same word at `SerializerAnalysisTest:220`.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
```diff
@@ -0,0 +1,444 @@
...
+	@Test
+	void bindsTheRosettaSerializerUnderBothTypes() {
+		// The gate lives in the serializer, so a language that binds Xtext's own has no gate at all.
+		//
+		// Both types, because Guice will happily build an unbound concrete class on demand and hand
+		// back a plain Serializer with no gate in it. That is not a hypothetical injection site:
+		// serializeToRegions and serializeReplacement are absent from ISerializer, so a caller wanting
+		// either has to ask for the concrete class, and XtextResourceFormatter does.
+		assertAll(
+				() -> assertInstanceOf(RosettaSerializer.class, injector.getInstance(ISerializer.class)),
```
SimonCockx, line 120: This assertion does not pin the `bindISerializer()` override. Deleting that override fails no test: `AbstractRosettaRuntimeModule:105` binds `ISerializer` to the concrete `Serializer`, and `configureSerializer` binds that to `RosettaSerializer`, so the chain resolves anyway — I ran it, 10/10 pass. Deleting `configureSerializer` is what fails, on line 120. Since the concern here is a future change quietly reopening the hole, worth a line saying which of the two bindings is load-bearing.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,228 @@
...
+	 * null — but the contract allows one, and {@code ContextFinder} reads the array expecting one, so
+	 * this skips rather than throws. Forcing is an optimisation of when work happens; it should not be
+	 * the thing that turns a tolerated null into a crash.
+	 */
+	private static void forceConstraintState(SerializationContextMap<IConstraint> constraints) {
+		for (SerializationContextMap.Entry<IConstraint> entry : constraints.values()) {
+			IConstraint constraint = entry.getValue();
+			// getAssignments() reaches getBody() as well, but only for a constraint that has at least
+			// one feature. Asking directly is what makes it closed for all of them.
+			constraint.getBody();
```
SimonCockx, line 185: Deleting this line fails no test — no constraint in the Rosetta grammar is featureless, so `getAssignments()` always reaches `getBody()` and `assertConstraintStateForced`'s body check never fires. Worth keeping as cover for a grammar that does produce one, but the comment claims coverage no test holds.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+
+	/**
+	 * Builds the analysis on the calling thread, returning once it is ready to be used from any
+	 * thread. A standalone consumer about to serialize from several threads can call this to pay the
+	 * cost up front, but does not have to: {@link #ensureWarm()} makes every serialization safe on its
+	 * own, so skipping this is slower on the first serialization and never less safe.
+	 */
+	public void warmUp() {
```
SimonCockx, line 125: Consider deleting `warmUp()`. It is `ensureWarm()` plus a timing log that reports "0 ms" whenever the analysis is already built, and the standalone consumer this javadoc describes has declined it: `RuneRuntime.warmUp(Injector)` in rosetta-components model-import says "do not swap this for it… serializing a real model warms the parser and the validation path as well. Swapping would warm strictly less." Moving the timing into `build()` would leave `ensureWarm` and `warmUpAsync` as the surface.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/RosettaSerializer.java
```diff
@@ -0,0 +1,54 @@
...
+ * <p>{@link Serializer#getIContext} is the one place every public entry point passes through before
+ * it touches a sequencer — {@code serialize} in each of its overloads, {@code serializeToRegions},
+ * and {@code serializeReplacement}. Warming up here therefore covers all of them, including
+ * {@code serializeToRegions}, which is not on {@link org.eclipse.xtext.serializer.ISerializer} and
+ * so cannot be covered by wrapping that interface. The two {@code protected serialize(context, …)}
+ * overloads take a context rather than deriving one and do not pass through here; they are reachable
+ * only from a subclass, which is why {@code SerializerAnalysisTest} pins that every public path
+ * still reaches this method.
+ *
+ * <p>Why this rather than guarding Xtext's analysis providers one at a time: the providers are not
```
SimonCockx: Just a preference: drop this paragraph. The same argument is at `RosettaRuntimeModule:139-141` and in the commit message, which is where a rejected alternative belongs, and the class is six lines of code under thirty-four of comment. No requirement.

rune-ide/src/test/java/com/regnosys/rosetta/ide/serializer/SerializerWarmUpServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+			@Override
+			public Map<String, Object> getProtocolToFactoryMap() {
+				return Map.of();
+			}
+		};
+	}
+
+	/**
+	 * A language whose only interesting property is which constraint provider it binds, and which
```
SimonCockx: "which constraint provider it binds" is stale — it is the serializer now.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
```diff
@@ -0,0 +1,444 @@
...
+	}
+
+	private static class CountingWarmUp extends SerializerAnalysisWarmUp {
+		private final SerializerAnalysisWarmUp delegate;
+		private final AtomicInteger calls = new AtomicInteger();
+
+		CountingWarmUp(SerializerAnalysisWarmUp delegate) {
+			// Safe because every method used here is overridden below and delegates; the injected
+			// fields are never touched.
+			super(null, null, null, null, null);
```
SimonCockx, line 447: More of an FYI: four doubles across this file and `SerializerWarmUpServiceTest` each pass `super(null, null, null, null, null)` with the same justification comment. Splitting the gate (`warm`/`published`/`ensureWarm`) from the builder would make each double a lambda instead, and the five nulls are the class telling you it does three jobs.

SimonCockx (commented):
The design landed faithfully and the blast radius is clean. One correctness bug inline, two lines that no test reaches, and four comment claims that do not hold against the Xtext source.

The bug is the `finally` at `SerializerAnalysisWarmUp:129`: a `build()` that throws still opens the gate permanently, which is the one failure that puts every later thread back on the half-built analysis.

Two things worth pulling out of the inline comments because they change the description rather than the code.

The description is stale. It still describes `SerializerAnalysisLock`, says "all four now build behind one `SerializerAnalysisLock`", carries the monitor's steady-state benchmark, and lists five things `SerializerAnalysisTest` pins. None of that is in the diff, and a reviewer reading it goes looking for a class that is not there. The commit message is current; the description has not caught up with it. The measurement table's middle rows describe configurations that are no longer what this change does, and neither the title nor the body mentions that the public `RosettaContextTypePDAProvider` is deleted.

`ModelImporterRunnerIsolationTest:79` in rosetta-products is `@Disabled` for exactly this defect — "Two formats importing for the first time concurrently corrupt Xtext grammar analysis; needs a sequential real-import warm-up" — and its javadoc independently records both crash symptoms this commit message names, at four failures in five runs. That is the strongest external evidence this works. Worth saying in the description whether it comes off in a follow-up and in what order, and heeding its sibling's warning: unwarmed failure is roughly four in five and the corruption persists for the life of the JVM, so one green run proves little.

## What I verified, and how

| Claim | How | Result |
|---|---|---|
| Deleting the gate is caught | mutation, 6 runs | `everyPublicSerializePathReachesTheGate` 6/6; concurrency test 5/6 |
| The three `indexByContext` provider calls are load-bearing | mutation | fails on syntactic PDAs and semantic NFAs |
| `forceConstraintState` is load-bearing | mutation | fails with a long list of unbuilt `.body`/`.features`/`.assignments` |
| `constraint.getBody()` is load-bearing | mutation | **survives**, 10/10 pass |
| `bindISerializer()` is load-bearing | mutation | **survives**; deleting `configureSerializer` is what fails |
| `warm` before `build()` | mutation | `StackOverflowError`, not deadlock |
| `published` needs `volatile` | mutation | survives, as expected — not reachable deterministically |
| Every test gets a cold injector | read + baseline | `RosettaIdeInjectorProvider.injector` is an instance field, so yes |
| The concrete `Serializer` is gated | runtime probe on a cold injector | `XtextResourceFormatter.serializer` = `RosettaSerializer`; serializing through it moved `published` false→true |
| Anything downstream overrides the bindings | grep + `gh search code` over 20 repos | none; six `RosettaRuntimeModule` subclasses, no serializer override, and the BSP's `overrideLanguageModules` is empty at all six call sites |
| Anything downstream reaches the analysis directly | same | no occurrences |
| Deleting `RosettaContextTypePDAProvider` breaks a consumer | same | no consumer outside rune-dsl |
| Xtext 2.41.0 still has the unsynchronized publication | read 2.41.0 sources | correct — four plain `HashMap` caches, `get`/`getFeatures` still assign-then-fill, `GrammarElementDeclarationOrder.get` unsynchronized |

Two limits on that coverage. `finos/rune-python-generator` is covered only on its default branch, so a serializer binding on a branch there would not show. And `ENABLE_SERIALIZER_WARM_UP` is set to true in all four rosetta-products environments, but only as a `-D` on the BSP subprocess (`BspVMArgs:94`, `bsp-proxy/src/bsp-launcher.ts:58`) — the backend JVM that runs the imports never receives it, so the flag does nothing for the case above.

The mutation matrix also makes the description's "tests that fail deterministically" more precise than it currently is: the concurrency test alone is about 83% per run, and it is `everyPublicSerializePathReachesTheGate` that makes gate deletion reliably visible in CI.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Now pinned by a test rather than only by reading. `oneSharedSerializerIsSafeFromSeveralThreads` shares one serializer across sixteen threads on a cold injector and asserts identical text: green 5 runs of 5, and not vacuous — with the gate removed it fails 2 of 3, once on `Cannot invoke "BiMap.inverse()" because "this.antlrNameToRule" is null` and once with a `ConcurrentModificationException`. So the javadoc now says the opposite of the note you asked for, with the reason: a serializer may be shared because it takes its sequencers from providers per call, and it is a sequencer that must not be.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Fixed in `e3f7c466e`. `published` is set only on a normal return and `warm` is cleared if `build()` throws, so the next caller rebuilds. `aFailedBuildLeavesTheGateShut` covers it, and two mutants kill it — restoring the `finally`, and clearing `published` without `warm`.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Correct, fixed. The upstream TODO now names `getFeatures` and `getAssignments` as the assign-then-fill pair, and the forcing comment gives the real reason for the body: the adapter it installs through `GrammarElementDeclarationOrder.get`.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Correct, fixed. The javadoc now says two are load-bearing in 2.38.0 — the syntactic automata and the constraints — and that all four are asked as cover for a later Xtext.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Dropped.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Fixed everywhere it appeared: the code comment, the `warm` field javadoc, the test comment and the commit message. Confirmed independently by mutation — 12.6s to `StackOverflowError`.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: Confirmed by mutation here too: deleting the override alone is 9/9 green, deleting `configureSerializer` is what fails. The comment now says which of the two closes the bypass and why the override is kept.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Agreed, and the comment no longer claims coverage. Kept as cover for a grammar that produces a featureless constraint, and because `getBody()` installs the declaration-order adapter itself.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/SerializerAnalysisWarmUp.java
SimonCockx: Considered and kept, with the reason now in the javadoc: it is the timed body `warmUpAsync` calls, and its elapsed-time log is the only view of what an eager warm-up cost.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/RosettaSerializer.java
SimonCockx: Kept, and the class is still comment-heavy. Its six lines are the whole gate, so the reasoning is what a reader arriving from the binding needs; happy to trim it if you would still rather.

rune-ide/src/test/java/com/regnosys/rosetta/ide/serializer/SerializerWarmUpServiceTest.java
SimonCockx: Fixed.

rune-ide/src/test/java/com/regnosys/rosetta/serializer/SerializerAnalysisTest.java
SimonCockx: Fair — the five nulls do say the class does three jobs. Not splitting it in this change, since the gate and the builder share the flags, but worth doing separately.

rune-lang/src/main/java/com/regnosys/rosetta/serializer/RosettaSerializer.java
SimonCockx: Done in `0e903b9c0`, and the same pass went over `SerializerAnalysisWarmUp` and `SerializerWarmUpService`. The paragraph is gone, the javadoc here is now two short paragraphs keeping only the funnel and why `serializeToRegions` forces the gate into `getIContext`. 62% comment down to 44%, 50% to 39%, 44% to 20%.

JayasriR (approved):

PR #1391, opened 2026-09-15
Let a tool configure a namespace it has generated but not yet written

SimonCockx:
A tool that generates a Rune namespace has to validate its model against a configuration for that namespace. There is nowhere to put one. The `rune-config.yml` is written after the model, so it does not carry the entry yet, and `RuneConfigurationHolder` holds a single value that every thread reads.

`RuneConfigurationHolder.overlay(UnaryOperator<RuneConfiguration>)` applies a function to the loaded configuration, on the calling thread only, until the returned scope closes. `overlayNamespaceConfigs(List<RuneNamespaceConfiguration>)` adds namespace entries through it, which is what a generator that has just produced a namespace needs:

```java
try (RuneConfigurationHolder.Scope scope = holder.overlayNamespaceConfigs(generated)) {
    validate(generatedModel);
}
```

Each entry is upserted by its id, so a namespace that is already configured is updated rather than competing with a second entry.

Nothing is written to disk, no other thread sees the overlay, and a `reload()` inside the scope is still picked up and still carries it.

One overlay at a time: opening a second on a thread that already has one throws. Two at once means either a scope that was never closed or work that has re-entered itself, and the second would validate against entries belonging to the first.

Nothing opens two overlays at once in model-import, the importer in rosetta-components that `overlay` was added for. That includes its suite that imports FpML (the XML standard for derivatives contracts), FIXML (the XML form of the FIX trading protocol) and CSV, generates Java from each and compiles it.

Three smaller things a tool that writes a configuration needs on the way out.

`FileBasedRuneConfigurationProvider` now reads each configuration file on its own, so a parse failure names the file. It reads the project's configuration and every dependency's, and one "unable to parse" over the lot left the reader to work out which was meant.

`RuneConfigurationService.readIfPresent(Path)` is empty for a file that is absent, blank, or nothing but comments. That is how a project asks for a configuration to be created rather than offering one to read, and `read` cannot make the distinction, because an empty file fails the same way a truncated one does.

`write(Path, ...)` names the file rather than leaving the caller with a bare "Is a directory".

Proven against that model-import, which was carrying its own version of all of this: an Xtext setup, a runtime module, a `FileBasedRuneConfigurationProvider` subclass, a singleton unioning the entries of every import currently open, and two test injector providers built on that module. Built against this branch, that change deletes seven classes and 490 lines net, and the model-import suites pass unchanged.

One fix came out of review rather than out of the feature. Reading a configuration and writing it back dropped every `generators.doNotPrune` entry: the accessor is `doNotPrune()` rather than `getDoNotPrune()`, so Jackson read it through the constructor's `@JsonProperty` and never wrote it out again. It has been that way since #1360 and nothing noticed, because nothing wrote a configuration back. ISO-20022 pins three entries there, and its configuration is one model-import is about to rewrite, so it is fixed here. That file now round-trips losing only a `dependencies:` line that has no value under it.

Each production file has tests here. `RuneConfigurationHolderTest` covers the overlay: it applies until its scope closes, it sees a `reload()` underneath it, a change that is not a namespace configuration applies the same way, a second overlay on the same thread is refused, another thread does not see it, closing an already-closed scope leaves a still-open one in force, and overlaid namespace configs upsert onto the configured ones. `RuneConfigurationServiceTest` covers `readIfPresent` (absent, empty, comment-only, valid, will-not-parse, not UTF-8) and a write that cannot happen naming the file it was for. The new `FileBasedRuneConfigurationProviderTest` covers the case reading each file on its own exists for: a dependency's configuration that will not parse fails naming that file, not the project's.

No issue number; this came out of that model-import work.

## Type of change

https://claude.ai/code/session_01HtCUD8o1GkT2pmeCsStVPQ


SimonCockx (commented):
Additive in rune-runtime: the old exception constructor stays and nothing is removed or renamed, so it is binary compatible. Three things inline, one of them optional.

|---|---|---|
| Nothing downstream matches the old "Unable to parse the Rosetta configuration." message | `gh search code` over finos, REGnosys, rosetta-models, plus a local grep of the bundle repos and rosetta-products | No matches |
| Subclasses of `FileBasedRuneConfigurationProvider` are not broken by the `read` split | Same search | Only rosetta-components model-import, which REGnosys/rosetta-components#961 deletes |
| `overlayNamespaceConfigs` upserts by id | Read `RuneConfiguration.Builder.addNamespaceConfig` | `removeIf` on id, then add |
| Scopes closed out of order leave an overlay behind | Ran a verbatim copy of `overlay` with outer closed before inner | `get()` still returns `Base+outer` after both are closed |
| `readIfPresent` on comment-only files with a BOM, a `---` marker, or CRLF line endings | Probe test in rune-runtime at 81d57b8 | BOM: throws "No content to map". The other two: empty |
| CI | Head 81d57b8 | Green on Ubuntu and Windows |

rune-lang/src/main/java/com/regnosys/rosetta/utils/RuneConfigurationHolder.java
```diff
@@ -49,6 +69,71 @@ public void reload() {
...
+	 * a pool sees the configuration without it.
+	 *
+	 * @param extra applied to the loaded configuration; must not be null
+	 * @return the scope to close, which is what takes the overlay away again
+	 */
+	public Scope overlay(UnaryOperator<RuneConfiguration> extra) {
+		Objects.requireNonNull(extra, "An overlay must be a function of the loaded configuration, not null.");
+		UnaryOperator<RuneConfiguration> previous = overlay.get();
+		overlay.set(previous == null ? extra : config -> extra.apply(previous.apply(config)));
+		return () -> {
```
SimonCockx: Closing outer before inner, or closing inner twice, silently leaves `Base+outer` on the thread, and on a pooled thread that carries into unrelated work. Throw in `close()` when the overlay in force isn't this scope's.

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
```diff
@@ -64,8 +67,55 @@ public RuneConfiguration readString(String yaml) throws IOException {
...
+			return Optional.ofNullable(mapper.readValue(path.toFile(), RuneConfiguration.class));
+		} catch (IOException e) {
+			String reason = e.getCause() == null ? e.getMessage() : e.getCause().getMessage();
+			throw new IOException("Cannot read the Rune configuration at " + path + ": " + reason, e);
+		}
+	}
+
+	private static boolean holdsContent(Path path) throws IOException {
+		try (Stream<String> lines = Files.lines(path)) {
+			return lines.map(String::trim).anyMatch(line -> !line.isEmpty() && !line.startsWith("#"));
```
SimonCockx: A comment-only file that starts with a UTF-8 BOM, which some Windows editors write, counts as content here and then throws "No content to map due to end-of-input". Strip a leading `\uFEFF` before the check.

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
```diff
@@ -64,8 +67,55 @@ public RuneConfiguration readString(String yaml) throws IOException {
...
+	 * The configuration at {@code path}, or empty where there is none to read: no file, or a file
+	 * holding only blank lines and comments.
+	 * <p>
+	 * A tool that updates a project's {@code rune-config.yml} needs that distinction. {@code read}
+	 * cannot make it, because a file with nothing in it fails the same way a truncated one does, and
+	 * an empty file is how a project asks for a configuration to be created rather than a corruption
+	 * to be reported. Anything else that will not parse still throws, and the message names the file.
+	 */
+	public Optional<RuneConfiguration> readIfPresent(Path path) throws IOException {
+		if (path == null || !Files.isRegularFile(path)) {
```
SimonCockx: Is a null path meant to read as "no file"? It hides a caller that forgot to pass one. I'd `requireNonNull` it instead.

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
```diff
@@ -64,8 +67,55 @@ public RuneConfiguration readString(String yaml) throws IOException {
...
+	 * the file system reports on its own -- "Is a directory", "Permission denied" -- identifies
+	 * nothing.
+	 */
 	public void write(Path path, RuneConfiguration configuration) throws IOException {
-		Files.write(path, writeString(configuration).getBytes(StandardCharsets.UTF_8));
+		byte[] yaml = writeString(configuration).getBytes(StandardCharsets.UTF_8);
+		try {
+			Files.write(path, yaml);
+		} catch (IOException e) {
+			String reason = e.getCause() == null ? e.getMessage() : e.getCause().getMessage();
```
SimonCockx: Just a preference: this `reason` line appears three times, here, in `readIfPresent` and in `FileBasedRuneConfigurationProvider.read`. A small shared helper would do. No requirement.

rune-lang/src/main/java/com/regnosys/rosetta/utils/RuneConfigurationHolder.java
SimonCockx: Fixed in ed48a0d3e: close() throws unless this scope is the overlay in force.

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
SimonCockx: Fixed in ed48a0d3e.

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
SimonCockx: Agreed, requireNonNull in ed48a0d3e. The command line has none to offer without -u, so it answers that itself now (REGnosys/rosetta-components#961).

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
SimonCockx: Taken: RuneConfigurationService.reason in ed48a0d3e.

rune-runtime/src/main/java/com/regnosys/rosetta/config/RuneConfigurationService.java
```diff
@@ -64,8 +70,77 @@ public RuneConfiguration readString(String yaml) throws IOException {
...
+		return trimmed.startsWith(BYTE_ORDER_MARK) ? trimmed.substring(1).trim() : trimmed;
+	}
+
+	/**
+	 * What a failure is worth telling the caller: the message of the cause where there is one, because
+	 * the wrapper's own message is usually the stack of types it came through rather than what went
+	 * wrong.
+	 */
+	public static String reason(Throwable failure) {
+		return failure.getCause() == null ? failure.getMessage() : failure.getCause().getMessage();
```
JayasriR: A cause with no message makes this return null, so the message ends in ": null" — a rune-config.yml with no `model:` key, where Jackson wraps the `Objects.requireNonNull(model)` NPE, gives this on both new paths:

```
FileBasedRuneConfigurationRuntimeException: Cannot read the Rune configuration at file:/tmp/rune-config.yml: null
IOException: Cannot read the Rune configuration at /tmp/rune-config.yml: null
```

Fall back to the cause's `toString()`, or to the wrapper's own message, when `getCause().getMessage()` is null.

SimonCockx: Fixed in 6bf176ace. Reproduced your case exactly: the cause is the bare NullPointerException from the required `model` field, so reason() falls back to the wrapper, which here is Jackson's "Cannot construct instance of RuneConfiguration, problem: ...", and then to the type name. I also gave that requireNonNull a message, so this particular file now reads "Cannot read the Rune configuration at ...: A Rune configuration needs a `model` section naming the model." rather than the Jackson wrapper text.

rune-lang/src/main/java/com/regnosys/rosetta/utils/RuneConfigurationHolder.java
```diff
@@ -49,6 +69,101 @@ public void reload() {
...
+	 * @param extra applied to the loaded configuration; must not be null
+	 * @return the scope to close, which is what takes the overlay away again
+	 */
+	public Scope overlay(UnaryOperator<RuneConfiguration> extra) {
+		Objects.requireNonNull(extra, "An overlay must be a function of the loaded configuration, not null.");
+		UnaryOperator<RuneConfiguration> previous = overlay.get();
+		UnaryOperator<RuneConfiguration> combined =
+				previous == null ? extra : config -> extra.apply(previous.apply(config));
+		overlay.set(combined);
+		return new NestedScope(previous, combined);
```
JayasriR: Just a preference, no requirement: putting the scope itself in the ThreadLocal drops the composition step and leaves one type instead of an interface with a single implementation.

```java
private final ThreadLocal<Scope> overlay = new ThreadLocal<>();

public RuneConfiguration get() {
    Scope active = overlay.get();
    return active == null ? loaded() : active.applyTo(loaded());
}

public Scope overlay(UnaryOperator<RuneConfiguration> extra) {
    Objects.requireNonNull(extra, "...");
    Scope scope = new Scope(extra, overlay.get());
    overlay.set(scope);
    return scope;
}

public final class Scope implements AutoCloseable {
    private final UnaryOperator<RuneConfiguration> extra;
    private final Scope enclosing;

    /** The enclosing overlays first, then this one, so an inner overlay sees the outer one's result. */
    private RuneConfiguration applyTo(RuneConfiguration config) {
        return extra.apply(enclosing == null ? config : enclosing.applyTo(config));
    }

    @Override
    public void close() {
        if (overlay.get() != this) { throw new IllegalStateException(...); }
        if (enclosing == null) { overlay.remove(); } else { overlay.set(enclosing); }
    }
}
```

It also removes something I had to prove while reading: at the outermost level `combined == extra`, so the guard relies on no caller-supplied function ever being in force twice — `new Scope(...)` is fresh by construction. Ran the eight tests here plus three-deep nesting and the same `UnaryOperator` overlaid at two levels against both versions: identical.

SimonCockx: Taken in f3259aae3, and it turned out to be more than a preference. The thing you had to prove while reading is exactly where the previous version was wrong: at the outermost level combined == extra, so two scopes over the same UnaryOperator instance are indistinguishable to the guard. Overlay one, close it, overlay the same instance again, and closing the spent scope passed the guard and took the live scope's overlay away. I added a test for it, which fails on the previous version and passes on yours.

JayasriR (approved):

PR #1393, opened 2026-09-16
Add description to schema

JayasriR:
Let a `schema` carry a description, as every other named element in the language already can.

`Schema` now extends `RosettaDefinable`, so this parses, formats and serializes:

```
schema fixml XML <"FIXML 5.0 SP2, as published by the FIX Trading Community.">
    [externalConfig]
```

The description sits after the format and before any annotations, following the same shape as `body` and `corpus`. It is optional, so every existing model parses unchanged.

The serialization test builds the schema in memory rather than parsing it. A tool that generates a Rune model constructs the declaration as an EMF object rather than writing text, so the description has to survive a serialization that has no node model to copy it from — that path is not exercised by a parse-and-format test.

## Type of change


website/docs/modelling-components/schema.mdx
```diff
@@ -28,7 +28,23 @@ The format refers to a value of the built-in `SerializationFormat` enumeration:
...
+schema fixml XML <"FIXML 5.0 SP2, as published by the FIX Trading Community.">
+```
+
+The description sits before any annotations:
+
+```
+schema fixml XML <"FIXML 5.0 SP2, as published by the FIX Trading Community.">
+    [externalConfig]
+```
+
```
SimonCockx: Not sure if this needs explicit documentation. This is already generally documented in a separate place.

JayasriR: yeah, agreed. good point, will remove it

SimonCockx (approved):

PR #1378, opened 2026-08-28
Convert RosettaGrammarUtil, EnumHelper, ModelGeneratorUtil, EnumGenerator and ConditionGenerator to Java

davidalk:
Stacked on #1398, which is stacked on #1400: the four commits from those PRs come first in this PR's diff. Two commits moved here from #1400 because they need this PR's Java generators: "Write enum value javadoc without splitting it into lines first" and "Indent the continuation lines of a condition definition spanning several lines". The last commit regenerates three regression fixtures, from #1398 and from this PR, that this PR's generators now render.

## Summary
- Migrates `RosettaGrammarUtil`, `EnumHelper` and `ModelGeneratorUtil` from Xtend to Java, then `EnumGenerator` and `ConditionGenerator` themselves, as part of the ongoing Xtend-to-Java migration (`.junie/xtend-migration-plan.md`).
- The three utility classes are plain translations (Java streams/StringBuilder in place of Xtend's `IterableExtensions`/`Function1`/`StringConcatenation`) — no generator output of their own to migrate to the fluent API.
- `ModelGeneratorUtil` builds javadoc comment text (for generated types, attributes, and doc references) line by line; the translation reproduces Xtend rich-string templates' "skip a blank line when its only placeholder was empty" behaviour with a small `terminateLine` helper, verified line-for-line against the existing `ModelGeneratorUtilTest` expectations.
- `EnumGenerator` extends `FluentRObjectJavaClassGenerator` and renders the enum class via `CodeRenderer`/`CodeWriter` instead of an Xtend template, delegating javadoc text to the now-Java `ModelGeneratorUtil`.
[...]

PR #1399, opened 2026-09-23
Escape model text in generated Java string literals and javadoc

SimonCockx:
Generated Java failed to compile when model text contained characters that mean something in Java source. Enum display names, type condition definitions, function condition definitions and the namespace `version` were copied into Java string literals without escaping, so a double quote or a backslash broke them. Documentation copied into javadoc broke on `*/`, which ends the comment, and on a backslash followed by `u`, as in `C:\users`, which Java reads as a unicode escape even inside a comment.

String literals are now escaped as Java. In javadoc, that backslash becomes `&#92;` and the slash of `*/` becomes `&#47;`, so the rendered javadoc reads the same. This covers every javadoc built by `ModelGeneratorUtil` (types, attributes, enums, enum values, function inputs and outputs, doc references, `@version`) and the generated `package-info`.

A new `escaping` regression model covers each of these cases. `EnumGeneratorTest`, `DataTypeConditionTest`, `FunctionGeneratorTest` and `ModelObjectGeneratorTest` also check that the generated display name, `DEFINITION` constant, condition failure message and `version` annotation value equal the model text. For the `package-info` case, the in-memory compiler in `rune-testing` now skips `package-info` when it loads compiled classes, because a `package-info` that only holds javadoc compiles to no class file.

Stacked on #1378, which is stacked on #1398 and #1400: only the top four commits belong to this PR: "Escape backslashes and quotes in generated enum display names and condition definitions", "Escape backslash-u and comment ends in generated javadoc", "Escape the model version in generated data type annotations" and "Escape function condition definitions in generated Java". The other files in this PR's diff belong to the PRs below it.

## Type of change


rune-lang/src/main/java/com/regnosys/rosetta/generator/java/enums/EnumGenerator.java
```diff
@@ -129,12 +129,13 @@ protected CodeRenderer generateClass(REnumType e, RJavaEnum javaEnum, String ver
 	private void renderEnumValue(CodeWriter out, RJavaEnumValue value) {
 		out.write(modelGeneratorUtil.javadoc(value.getEObject()));
 		String displayName = value.getDisplayName();
+		String displayNameLiteral = displayName != null ? "\"" + StringEscapeUtils.escapeJava(displayName) + "\"" : null;
```
SimonCockx: Can we use `JavaLiteral.STRING(displayName)` here, with `JavaLiteral.NULL` for the missing case? It already escapes, so this site can't forget to.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/function/FunctionGenerator.java
```diff
@@ -881,7 +883,7 @@ private JavaExpression lhsFeature(JavaExpression receiver, RosettaFeature f) {
 	private void renderContributeCondition(CodeWriter out, Condition condition, GeneratedIdentifier conditionValidator, JavaStatementScope scope) {
 		JavaStatementBuilder conditionBody = expressionGenerator.javaCode(condition.getExpression(), typeUtil.COMPARISON_RESULT, scope.lambdaScope());
 		out.writeln(conditionValidator, ".validate(() -> ", conditionBody.toLambdaBody(), ",");
-		out.writeln("    \"", condition.getDefinition(), "\");");
+		out.writeln("    \"", StringEscapeUtils.escapeJava(condition.getDefinition()), "\");");
```
SimonCockx: Same here: `JavaLiteral.STRING(...)`, which also removes the new import. Guard null with `Objects.requireNonNullElse(condition.getDefinition(), "")`, because `STRING(null)` renders `"null"`.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/object/ModelObjectGenerator.xtend
```diff
@@ -87,10 +88,11 @@ class ModelObjectGenerator extends RObjectJavaClassGenerator<RDataType, JavaPojo
 		val builderImplClass = javaType.toBuilderImplClass
 		val builderImplScope = pojoScope.createNestedClassScopeAndRegisterIdentifier(builderImplClass)
 		val modelShortName = javaType.packageName.first
+		val escapedVersion = StringEscapeUtils.escapeJava(javaType.version)
```
SimonCockx: `«JavaLiteral.STRING(javaType.version)»` in both annotations would replace this local and the hand-written quotes.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/ModelGeneratorUtil.java
```diff
@@ -133,13 +154,13 @@ private String javadocDocRef(List<RosettaDocReference> references) {
...
 							.append(" \"")
-							.append(HtmlEscapers.htmlEscaper().escape(segment.getSegmentRef()))
+							.append(escape(segment.getSegmentRef()))
 							.append("\"");
 				}
 				terminateLine(out);
 				out.append(" *");
 				terminateLine(out);
-				out.append(" * Provision ").append(reference.getProvision());
+				out.append(" * Provision ").append(reference.getProvision() != null ? escapeComment(reference.getProvision()) : null);
```
SimonCockx: A docReference without `provision` (optional in the grammar) renders ` * Provision null` here, because `append(null)` appends `"null"`. The Xtend version printed nothing. This came in with #1378 (I read the code but did not run it). Since this line is touched: let `escapeComment` return `""` for null and drop the ternaries here and on line 73.

SimonCockx (commented):
The escaping itself looks right. I found no other place where model text reaches generated Java unescaped. Of the 16 mutations I ran against the change (each one reverts a single escape), 15 fail a test.

One cause sits behind most of the inline comments. Each call site pastes model text into a literal or a comment by hand and picks its own escaper: `escapeJava` with hand-added quotes in three places, `RosettaGrammarUtil.escapeLines`, `escape`, and `escapeComment`. That is how the version, the provision and the package-info text came to get comment escaping but no HTML escaping. `JavaLiteral.STRING` already renders an escaped Java string literal. Using it at every literal site means a new site can't forget to escape. The same bug is still open outside this diff: from reading the code (not run), `FunctionGenerator.java:214` writes `configPath` raw, so a Windows path such as `a\b.json` in the config would silently turn into a backspace.

About 1,070 of the 1,192 added lines are expected regression output. The three fixtures flagged inline account for about 870 of them. According to the mutation results, every mutation they catch is also caught by a smaller test.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/ModelGeneratorUtil.java
```diff
@@ -70,23 +70,44 @@ public String emptyJavadocWithVersion(String version) {
...
 		out.append(" */");
 		terminateLine(out);
 		return out.toString();
 	}
 
+	/**
+	 * Turns model text into javadoc text: HTML-escapes it, so the text renders as written, and
+	 * makes it safe inside a comment (see {@link #escapeComment(String)}).
+	 */
 	public String escape(String definition) {
```
SimonCockx: Just a preference: rename the parameter to `text`. It also receives display names and segment refs. No requirement.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/RosettaGrammarUtil.java
```diff
@@ -43,10 +43,18 @@ public static String quoteForCodeWriter(String text) {
 		return "\"" + escapeLines(text).replace("\n", "\\n\" +\n\"") + "\"";
 	}
 
+	/**
+	 * Escapes model text for a Java string literal and normalises its line breaks to single
```
SimonCockx: "Normalises its line breaks to single `\n`s" isn't accurate: `replace` makes one pass, so three newlines become two. Can we drop that clause and the last sentence, and keep only the part about why backslashes go before quotes?

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/RosettaGrammarUtil.java
```diff
@@ -43,10 +43,18 @@ public static String quoteForCodeWriter(String text) {
...
+	 * {@code \n}s. Backslashes are escaped before quotes, so the backslash added in front of a
+	 * quote is not doubled again. Doubling every backslash also stops the Java compiler from
+	 * reading a backslash followed by {@code u} as a unicode escape.
+	 */
 	private static String escapeLines(String text) {
 		return text.trim()
+				.replace("\\", "\\\\")
 				.replace("\"", "\\\"")
 				.replace("\r\n", "\n")
+				.replace("\r", "\n")
```
SimonCockx: No test fails when this line is removed (I tried), and it isn't part of the bug the PR describes. Drop it, or add a fixture with a bare CR.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/RosettaGrammarUtil.java
```diff
@@ -43,10 +43,18 @@ public static String quoteForCodeWriter(String text) {
 		return "\"" + escapeLines(text).replace("\n", "\\n\" +\n\"") + "\"";
 	}
 
+	/**
+	 * Escapes model text for a Java string literal and normalises its line breaks to single
+	 * {@code \n}s. Backslashes are escaped before quotes, so the backslash added in front of a
+	 * quote is not doubled again. Doubling every backslash also stops the Java compiler from
+	 * reading a backslash followed by {@code u} as a unicode escape.
+	 */
 	private static String escapeLines(String text) {
```
SimonCockx: More of an FYI: this is a second hand-written Java escaper. Splitting on `\n` first and applying `StringEscapeUtils.escapeJava` to each line would leave one. Its output differs for tabs, control characters and non-ASCII.

rune-integration-tests/src/test/resources/generation-regression-tests/condition/model/condition.rosetta
```diff
@@ -29,6 +29,12 @@ func IsAllowed:
     set result:
         input -> val <> "forbidden"
 
+type Escaped:
```
SimonCockx: Optional, to shrink the PR: this fixture adds about 430 lines of expected output. `DataTypeConditionTest`'s new test catches the same mutation (the missing backslash escape) and asserts the value.

rune-integration-tests/src/test/resources/generation-regression-tests/enum/model/enums.rosetta
```diff
@@ -19,3 +19,6 @@ enum Documented: <"An enumeration with documentation.">
 
 enum Extended extends Documented: <"An enumeration that extends another one.">
     FIFTH <"An extra value.">
+
+enum Escaped: <"An enumeration whose display name needs escaping in Java.">
```
SimonCockx: Optional, same reason: `EnumGeneratorTest`'s new test catches every mutation this one does.

rune-integration-tests/src/test/resources/generation-regression-tests/escaping/model/escaping.rosetta
```diff
@@ -0,0 +1,28 @@
+namespace test.escaping : <"A namespace stored in C:\\users, which */ ends a comment.">
+version "1.0 \"beta\" C:\\users"
+
+body Organisation Org1 <"A body in C:\\users */">
+corpus Agreement Org1 "Agreement in C:\\users */" Agr1 <"A corpus in C:\\users */">
+
+segment name
+
+type Documented: <"A type stored in C:\\users, which */ ends a comment.">
```
SimonCockx: Optional: this type generates 5 expected files (about 380 lines). Its javadoc goes through the same `ModelGeneratorUtil` code as `DocumentedEnum`. `ModelObjectGeneratorTest` catches the annotation `version` and the Meta javadoc. By the mutation results nothing is lost, though I haven't re-run the tests without it.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/enums/EnumGenerator.java
SimonCockx: Done in 08e29284e.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/function/FunctionGenerator.java
SimonCockx: Done in 08e29284e, which also renders `configPath` on line 214 with `JavaLiteral.STRING`. No test has a backslash in a `configPath`.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/object/ModelObjectGenerator.xtend
SimonCockx: Done in 08e29284e.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/ModelGeneratorUtil.java
SimonCockx: Fixed in 186c68caf.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/ModelGeneratorUtil.java
SimonCockx: Done in 186c68caf.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/RosettaGrammarUtil.java
SimonCockx: Done in 70a0722b3.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/RosettaGrammarUtil.java
SimonCockx: Dropped in 70a0722b3.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/RosettaGrammarUtil.java
SimonCockx: I'd rather leave that to a separate PR: it changes every generated condition `DEFINITION` that contains a tab or non-ASCII text, and `quote` also feeds the C# generator in rosetta-code-generators.

rune-integration-tests/src/test/resources/generation-regression-tests/condition/model/condition.rosetta
SimonCockx: Removed in 4315c88d4.

rune-integration-tests/src/test/resources/generation-regression-tests/enum/model/enums.rosetta
SimonCockx: Removed in 4315c88d4.

rune-integration-tests/src/test/resources/generation-regression-tests/escaping/model/escaping.rosetta
SimonCockx: Removed in 4315c88d4. Without it, removing either replacement in `escapeComment` still fails 6 tests.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/RosettaGrammarUtil.java
SimonCockx: On second thought, done in 98d4eed8f: `ConditionGenerator` now writes one `JavaLiteral.STRING` per line and no longer uses `RosettaGrammarUtil`, which is deprecated and only kept for the C# generator. 67cd17234 does the same for the label provider and format validator literals, so `JavaLiteral` is the only caller of `escapeJava` left in rune-lang.

davidalk (approved):

PR #1400, opened 2026-09-23
Indent every line of multi-line text written to a CodeWriter

SimonCockx:
Text containing line breaks written to a `CodeWriter` lost its indentation after the first line: the writer only indented at the start of a line, so a multi-line javadoc written inside a class body came out at column 0. Every caller had to split such text into lines itself, and the enum-value javadoc in #1378 got this wrong.

`CodeWriter.write` now treats each `\n` in the text like `newline()` and indents every line that has content, so callers can write multi-line text in one call. Blank lines stay empty, and a `\r` before a `\n` is dropped. The Xtend bridge writer, `TargetStringConcatenationCodeWriter`, follows the same rules. Text without a line break costs one extra `indexOf('\n')`. This changes no generated output on `main`.

#1398, #1378 and #1399 are stacked on this PR, in that order.

## Type of change


rune-generator-api/src/main/java/com/regnosys/rosetta/codegen/support/AbstractCodeWriter.java
```diff
@@ -55,11 +56,47 @@ public void write(Object object) {
             renderer.render(this);
             return;
         }
+        String text = object.toString();
```
SimonCockx: Just a suggestion: the single-line branch and `writeLines` fold into one loop, which drops the `firstLineEnd` parameter and the second copy of the indent block:

```java
String text = object.toString();
int lineStart = 0;
int lineEnd;
while ((lineEnd = text.indexOf('\n', lineStart)) >= 0) {
    int contentEnd = lineEnd > lineStart && text.charAt(lineEnd - 1) == '\r' ? lineEnd - 1 : lineEnd;
    writeLineContent(text, lineStart, contentEnd);
    newline();
    lineStart = lineEnd + 1;
}
writeLineContent(text, lineStart, text.length());
```

Only difference: `write("")` at the start of a line no longer writes the indentation. All integration and generator-api tests pass with it.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriter.java
```diff
@@ -54,13 +54,50 @@ public void write(Object object) {
...
 		target.append(object);
 	}
 
+	/**
+	 * Writes text containing at least one line feed, of which {@code firstLineEnd}
+	 * is the index of the first, following the rules of {@link CodeWriter#write(Object)}.
+	 * Only strings are split: other objects are handed to the target as they are,
+	 * so that it can handle them itself.
+	 */
+	private void writeLines(String text, int firstLineEnd) {
```
SimonCockx: The same fold applies here. Could we also add a migration note saying this intentionally duplicates `AbstractCodeWriter`'s line handling and goes with the bridge, like `RecordingCodeWriter` has?

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriter.java
```diff
@@ -54,13 +54,50 @@ public void write(Object object) {
...
 			target.append(INDENT.repeat(indent));
 			atStartOfLine = false;
 		}
 		target.append(object);
 	}
 
+	/**
+	 * Writes text containing at least one line feed, of which {@code firstLineEnd}
+	 * is the index of the first, following the rules of {@link CodeWriter#write(Object)}.
+	 * Only strings are split: other objects are handed to the target as they are,
```
SimonCockx: Can we add a test for this? Splitting every object's `toString()` instead still passes all four tests in `TargetStringConcatenationCodeWriterTest`.

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriterTest.java
```diff
@@ -0,0 +1,76 @@
...
+	@Test
+	void emptyLinesStayEmptyAndCarriageReturnsAreDropped() {
+		assertEquals("    first\n\n    second", render(out -> {
+			out.indent();
+			out.write("first\r\n\r\nsecond");
+		}));
+	}
+
+	@Test
+	void textAfterLastLineFeedLeavesWriterMidLine() {
```
SimonCockx: Add a case with text starting with `\n`, e.g. `write("a"); write("\nb")`. Removing the `lineEnd > lineStart` guard, or changing `lineEnd >= 0` to `> 0`, both pass this class today.

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/util/RecordingCodeWriterTest.java
```diff
@@ -87,6 +87,20 @@ void indentationIsAppliedByReplayTarget() {
 				}""", replayToString());
 	}
 
+	@Test
+	void multiLineTextIsIndentedByReplayTarget() {
```
SimonCockx: Can this one go? `RecordingCodeWriter` passes text through unchanged, so this only re-tests `StringCodeWriter`, which `AbstractCodeWriterTest` already covers.

SimonCockx (commented):
No correctness issues found; comments are simplification and test gaps. I checked the tests by hand-mutating the new code:

| Mutation | `AbstractCodeWriterTest` | `TargetStringConcatenationCodeWriterTest` |
|---|---|---|
| keep `\r` before `\n`; indent empty lines; indent mid-line; drop text after last `\n`; no start-of-line reset; never clear `atStartOfLine`; always drop char before `\n` | killed | killed |
| drop `lineEnd > lineStart` guard | killed | **survived** |
| `lineEnd >= 0` → `> 0` | killed | **survived** |
| split non-`String` objects too | n/a | **survived** |

On cost: a synthetic 200-method class written through `StringCodeWriter` goes from ~112 to ~125 µs per file (plain timing loop, not JMH), so the extra `indexOf` is ~10% of the writer but negligible next to the recording pass.

rune-generator-api/src/test/java/com/regnosys/rosetta/codegen/support/AbstractCodeWriterTest.java
```diff
@@ -113,4 +113,96 @@ void testCustomNewline() {
...
+    @Test
+    void testMultiLineTextAfterMidLineWriteContinuesTheLine() {
+        out.indent();
+        out.write("int x = ");
+        out.write("1 +\n2;");
+        assertEquals("    int x = 1 +\n    2;", out.toString());
+    }
+
+    @Test
+    void testMultiLineTextInsideIndented() {
```
SimonCockx: No requirement: this has the same input and indent as `testEmptyLinesInMultiLineTextStayEmpty` and catches no mutant that one misses.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriter.java
```diff
@@ -54,13 +54,50 @@ public void write(Object object) {
 			renderer.render(this);
 			return;
 		}
+		if (object instanceof String text) {
```
davidalk: Widen this to `CharSequence`. An Xtend rich string is a `StringConcatenation`, not a `String`, so every `def foo() '''…'''` — the most common multi-line value in this repo — still skips the split here, and the two writers disagree on exactly the input the new `CodeWriter.write` javadoc describes generically as "text".

Reproduced on this branch, writing the same `"/**\n * doc\n */\n"` at indent 1:

```
bridge, String       = "    /**\n     * doc\n     */\n    class C {}"
bridge, rich string  = "    /**\n * doc\n */\nclass C {}"              <- unchanged from main
AbstractCodeWriter   = "    /**\n     * doc\n     */\n    class C {}"  (both forms)
```

No live caller hits it today — `FunctionGenerator:519` writes `modelGeneratorUtil.javadoc(…)` (a rich string), but through `RecordingCodeWriter`, which calls `toString()` before recording — so this is latent rather than a live bug.

`CharSequence` is safe as the wider test: nothing that needs the target's own machinery is one. `GeneratedIdentifier`, `TargetLanguageRepresentation`, `JavaClass` and `JavaType` are not, and `CharSequence` appears nowhere in `generator/java/types`.

davidalk (changes requested):
The core fix is right and the 12 new cases cover the edges well — empty lines, CR, configured newline, mid-line continuation, nesting. One gap: the bridge splits only `String`, so Xtend rich strings still come out unindented, which I reproduced against `75a07e50`. Details inline.

rune-generator-api/src/main/java/com/regnosys/rosetta/codegen/support/AbstractCodeWriter.java
SimonCockx: Fixed in ffa56f993.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriter.java
SimonCockx: Fixed in ffa56f993, including the migration note.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriter.java
SimonCockx: Added in ffa56f993.

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriterTest.java
SimonCockx: Added in ffa56f993.

rune-integration-tests/src/test/java/com/regnosys/rosetta/generator/java/util/RecordingCodeWriterTest.java
SimonCockx: Removed in ffa56f993.

rune-generator-api/src/test/java/com/regnosys/rosetta/codegen/support/AbstractCodeWriterTest.java
SimonCockx: Removed in ffa56f993.

rune-lang/src/main/java/com/regnosys/rosetta/generator/java/util/TargetStringConcatenationCodeWriter.java
SimonCockx: Fixed in e7ffe01dc, with your rich-string case as a test. `TargetLanguageStringConcatenation` is also a `CharSequence`, but it is only ever a top-level buffer, never a value passed to `write`.

davidalk (approved):

